PR #14, opened 2019-10-15
sty: pep8 complaints

mgxd:
some pep8 fixes - reformatted similar to black.


.pep8speaks.yml
```diff
@@ -5,7 +5,10 @@ scanner:
 pycodestyle:  # Same as scanner.linter value. Other option is flake8
     max-line-length: 88  # Follow black's line length
     ignore:  # Errors and warnings to ignore
+        - E266  # Too many leading '#' for block comments
```
oesteban: I'd add all the relevant ignores to the `[flakes8]` section of the setup.cfg. Please check nibabel's (https://github.com/nipy/nibabel/blob/debf173903b949f21ff57369f99388768263a7bc/setup.cfg#L76-L83) since we will end up merging this code in there.

PR #15, opened 2019-10-15
tst: skip unfinished implementation tests

mgxd:
Preceeds #14


nitransforms/nonlinear.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, field, reference=None):
         self._field = field.get_data()
 
         ndim = self._field.ndim - 1
-        if self._field.shape[:-1] != ndim:
```
mgxd: @oesteban is this the intended behavior?

nitransforms/nonlinear.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, field, reference=None):
...
+        if len(self._field.shape[:-1]) != ndim:
```
mgxd: @oesteban is this the intended behavior?

oesteban: I'm guessing so.

oesteban: I'm guessing so :)

PR #17, opened 2019-10-18
ENH: FreeSurfer LTA file support

mgxd:
This PR builds off nipy/nibabel#565 to support reading/writing transforms to the LTA format.


nitransforms/patched.py
```diff
@@ -22,7 +22,7 @@ def obliquity(affine):
         The *obliquity* of each axis with respect to the cardinal axes, in radians.
     """
     vs = voxel_sizes(affine)
-    best_cosines = np.abs((affine[:-1, :-1] / vs).max(axis=1))
+    best_cosines = np.abs(affine[:-1, :-1] / vs).max(axis=1)
```
oesteban: The whole obliquity thing has been merged into nibabel. Can you rebase after I update master?

nitransforms/io.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+class LinearTransformArray(StringBasedStruct):
+    template_dtype = np.dtype([
+        ('type', 'i4'),
+        ('nxforms', 'i4'),
+        ('subject', 'U1024'),
+        ('fscale', 'f4')])
+    dtype = template_dtype
+    _xforms = None
```
oesteban: I would not allow this to contain VOX2VOX. Meaning, if ``transform_code`` is 0, then the transforms are decomposed and the RAS2RAS extracted. If that is not possible because moving and/or reference VOX2RAS are missing, then raise an error.

That said, I'd be fine with a `NotImplementedError` when transform code is 0.

mgxd: since this builds off @effigies implementation, we should rope him in here. 

I assumed the scope of his LTA implementation is greater than just nitransforms' use-case, so we may want to still support vox2vox as a valid matrix. however, totally agree we should catch that case within the transforms module, and coerce it into ras2ras.

effigies: We should definitely permit reading/writing non-RAS2RAS, even if we only ever store RAS2RAS. I vaguely recall I might have intended to store the incoming transform, so that a load-save round trip would not change the contents, and only convert at need, but don't feel bound by this.

mgxd: would the `mean` / `sigma` change if we convert the matrix between transform types?

nitransforms/io.py
```diff
@@ -0,0 +1,263 @@
...
+                sa[key] = val.reshape(sa[key].shape) if val.size else ''
+
+        assert len(lta._xforms) == sa['nxforms']
+        return lta
+
+    @classmethod
+    def from_fileobj(klass, fileobj, check=True):
+        return klass.from_string(fileobj.read())
+
+    def as_type(self, target):
```
effigies: For me, `as_type()` reads that it would return a modified version of the data without modifying the source data.

```suggestion
    def set_type(self, target):
```

nitransforms/io.py
oesteban: Please note I'm not saying we should only support `LINEAR_RAS_TO_RAS`, I'm saying we should not write (just write) `LINEAR_VOX_TO_VOX`.

VOX2VOX is a legacy method that only makes sense in the context of the early development of image registration. Why (and who) anyone would like to write VOX2VOX? There's literally nothing VOX2VOX can do that cannot be done with RAS2RAS.

judging by https://github.com/freesurfer/freesurfer/blob/d5ff65ce78fee3ef296cc0b4027360ba6f9721f1/utils/transform.cpp#L823, I don't think `sigma` or `mean` should change with RAS2RAS.

oesteban: I would be more compelled by some demonstration of better numerical stability or precision of VOX2VOX, but I would actually guess that's also going to work in the opposite way.

oesteban: After checking further, it seems mean and sigma are not necessary for RAS2RAS at all (https://github.com/freesurfer/freesurfer/blob/b156d5aee6df2c7027ea45d3824813f8dcc536ef/lta_convert/lta_convert.cpp#L337).

nitransforms/linear.py
```diff
@@ -326,6 +342,13 @@ def load(filename, fmt='X5', reference=None):
     # elif fmt.lower() == 'afni':
     #     parameters = LPS.dot(self.matrix.dot(LPS))
     #     parameters = parameters[:3, :].reshape(-1).tolist()
+    elif fmt.lower() in ('fs', 'lta'):
+        with open(filename) as ltafile:
+            lta = LinearTransformArray.from_fileobj(ltafile)
+        assert lta['nxforms'] == 1  # ever have multiple transforms?
```
oesteban: Yes, we should be able to have multiple transforms (i.e., nxforms does not need to be 1)

nitransforms/linear.py
```diff
@@ -326,6 +342,13 @@ def load(filename, fmt='X5', reference=None):
...
+        if lta['type'] != 1:
+            lta.as_type(1)
+        matrix = lta['xforms'][0]['m_L']
```
oesteban: matrix is of size N x (D + 1) x (D + 1), where N is the number of transforms and D the dimension (i.e., D 
belongs to {2, 3})

Issue #22, opened 2019-10-21
Implement ``__call__`` of transforms


oesteban:
Example:

```
>>> xfm = Affine(matrix=mat)
>>> xfm((2.3, -6.5, 10.7))
[(12.3, 4.5, 0.7)]
```

i.e., it just calls `map(coords)`

PR #24, opened 2019-10-22
ENH: Uber-refactor of code style, method names, etc.

oesteban:
Includes a draft implementation of #22.

### DIFF
--- a/nitransforms/base.py
+++ b/nitransforms/base.py
@@ -15,13 +15,14 @@ from scipy import ndimage as ndi
 EQUALITY_TOL = 1e-5
 
 
-class ImageSpace(object):
+class ImageGrid(object):
     """Class to represent spaces of gridded data (images)."""
 
     __slots__ = ['_affine', '_shape', '_ndim', '_ndindex', '_coords', '_nvox',
                  '_inverse']
 
     def __init__(self, image):
+        """Create a gridded sampling reference."""
         self._affine = image.affine
         self._shape = image.shape
         self._ndim = len(image.shape)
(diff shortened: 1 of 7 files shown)


nitransforms/base.py
```diff
@@ -150,18 +160,25 @@ def resample(self, moving, order=3, mode='constant', cval=0.0, prefilter=True,
...
+
         """
         moving_data = np.asanyarray(moving.dataobj)
         if output_dtype is None:
             output_dtype = moving_data.dtype
 
+        moving_grid = ImageGrid(moving)
+
+        def _map_indexes(ijk):
+            return moving_grid.inverse.dot(self.map(self.reference.affine.dot(ijk)))
```
oesteban: I think we can reduce this down to defining this concatenation, where the affines of the moving and reference grids are applied on input indexes.

nitransforms/nonlinear.py
```diff
@@ -51,111 +51,46 @@ def __init__(self, field, reference=None):
...
-                             0, -1)
-
-        return coordinates + deltas
+        """
+        if inverse is True:
+            raise NotImplementedError
+        ijk = self.reference.index(x)
+        indexes = np.round(ijk).astype('int')
+        if np.any(np.abs(ijk - indexes) > 0.05):
+            print('Some coordinates are off-grid of the displacements field.',
```
mgxd: why not use `warnings` here?

nitransforms/nonlinear.py
```diff
@@ -51,111 +51,46 @@ def __init__(self, field, reference=None):
...
+                  file=sys.stderr)
+        indexes = tuple([tuple(i) for i in indexes.T])
```
mgxd:
```suggestion
        indexes = tuple(tuple(i) for i in indexes.T)
```

nitransforms/nonlinear.py
oesteban: Will do. Thanks!

PR #30, opened 2019-10-24
ENH: Rename ``resample()`` with ``apply()``

### LINKED ISSUE #7: Rename the `resample` function to `apply` (or similar)
Suggestions are welcome. The api I am thinking is like:

```
resampled = xfm.apply(moving)
```
where img can be a path or a nifti image, and the internal reference of the transform is used. The method should have an optional `reference` argument:
```
resampled = xfm.apply(moving, reference=ref)
```
where ref is a path or object pointing to a nifti/gifti/cifti, or any data structure (image, surface, pointset) supported by nibabel. Here ref would replace the internal reference of the transform.

Then, additional optional arguments will allow changing the resampling settings.

Limitations of the implementation:

- moving cannot be a surface/sphere/pointset:
  - resampling a surface onto a volume might not make sense
  - resampling a surface onto another surface would be great to have, but I doubt current implementation can support this use case.
  - pointset to volume/surface might not make sense either (and seems like an ill-posed problem)

EXAMPLE:
```Python
xfm.apply('sub-01_task-mixedgamblestask_run-01_bold.nii.gz', 
          reference=fslr).to_filename(
    'sub-01_task-mixedgamblestask_run-01_bold.dtseries.nii')
```
where fslr is a cifti file containing grayordinates, and xfm is a composite of all transforms from fslr to the bold dataset (including head motion correction, for instance).

oesteban:
This PR introduces the possibility of applying transforms to images when the reference is not an image.


nitransforms/base.py
```diff
@@ -131,16 +184,19 @@ def ndim(self):
...
-        Resample the moving image in reference space.
+        Apply a transformation to an image, resampling on the reference spatial object.
 
         Parameters
         ----------
-        moving : `spatialimage`
+        spatialimage : `spatialimage`
             The image object containing the data to be resampled in reference
             space
+        reference : spatial object
```
mgxd:
```suggestion
        reference : spatial object, optional
```

nitransforms/base.py
```diff
@@ -160,21 +216,27 @@ def resample(self, moving, order=3, mode='constant', cval=0.0, prefilter=True,
...
+
+        _ref = self.reference if reference is None \
+            else SpatialReference.factory(reference)
 
-        moving_data = np.asanyarray(moving.dataobj)
-        output_dtype = output_dtype or moving_data.dtype
-        targets = ImageGrid(moving).index(
-            _as_homogeneous(self.map(self.reference.ndcoords.T),
-                            dim=self.reference.ndim))
+        if isinstance(spatialimage, str):
```
mgxd: since we're allowing reference to be a `Path`
```suggestion
        if isinstance(spatialimage, (str, Path)):
```

oesteban:
Thanks very much!

PR #36, opened 2019-10-28
TST+FIX: LTA conversions

mgxd:


nitransforms/tests/data/v2v.lta
```diff
@@ -0,0 +1,30 @@
+# transform file v2v.lta
```
oesteban: How did you calculate this file?

mgxd: I used `mri_robust_register` on the the first volume of two bold runs:

```
$ mri_robust_register --mov run1.nii.gz --dst run2.nii.gz --lta r2r.lta
$ mri_robust_register --mov run1.nii.gz --dst run2.nii.gz --lta v2v.lta --vox2vox
```

oesteban:
I think it won't change your tests, but I think this line has source and destination swapped:
https://github.com/poldracklab/nitransforms/blob/45a0a2fe4268714b0f7237bab65cb084d806fa68/nitransforms/io/lta.py#L243

That line would work okay if `xform['m_L']` were the ras2ras matrix and you wanted to calculate the vox2vox matrix from it. However, we want to do the opposite:
* Let R be a matrix representing a ras2ras transform and V the matrix representing the equivalent vox2vox transform.
* Let Asrc and Adst be the affines of source and destination, respectively.

Then, R = inv(Adst) . V . Asrc.

So, if we want to extract V from R, we need to multiply the inverses of each side:
```Python
            M = dst.as_affine().dot(xform['m_L'].dot(np.linalg.inv(src.as_affine())))
```

mgxd:
yeah, i was just looking at how LTA convert implements and you got it
https://github.com/freesurfer/freesurfer/blob/933dc7794c15243e7f0370e912c6e877b87eda66/utils/transform.cpp#L3707

Even changing the algorithm, I'm still not sure why it is not working in my example...

oesteban:
Can you update the line with my suggestion?

mgxd:
Okay - I decided to use the `affine-RAS.fs.lta` as the baseline (RAS2RAS), and convert it to VOX2VOX using `lta_convert`. There is still a _slight_ discrepancy, albeit much smaller than before.

```
In [3]: v2v_m                                                                                                                                                                                                                               
Out[3]: 
array([[ 9.9999899e-01, -9.9999935e-04,  9.9999981e-04,  3.9999959e+00],
       [ 1.4049364e-03,  6.2160885e-01, -7.8332651e-01,  1.9999948e+00],
       [ 1.6171722e-04,  7.8332722e-01,  6.2160969e-01, -9.9998468e-01],
       [ 0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  1.0000000e+00]],
      dtype=float32)

In [4]: r2r_m                                                                                                                                                                                                                               
Out[4]: 
array([[ 9.9999899e-01, -9.9999935e-04,  9.9999981e-04,  4.0000000e+00],
       [ 1.4049363e-03,  6.2160885e-01, -7.8332651e-01,  2.0000000e+00],
       [ 1.6171722e-04,  7.8332716e-01,  6.2160963e-01, -1.0000000e+00],
       [ 0.0000000e+00,  0.0000000e+00,  0.0000000e+00,  1.0000000e+00]],
      dtype=float32)
```

nitransforms/io/lta.py
```diff
@@ -191,6 +191,13 @@ def from_string(klass, string):
         lta = klass()
         sa = lta.structarr
         lines = string.splitlines()
+        while lines:
+            if not lines[0].startswith('type'):
+                lines.pop(0)
+                continue
+            break
+        if not lines:
+            raise Exception("Invalid LTA format")
```
oesteban:
```suggestion
            raise TransformFileError("Invalid LTA format")
```
(and please add import on top)

nitransforms/io/lta.py
```diff
@@ -92,7 +92,7 @@ def from_string(klass, string):
 
 class LinearTransform(StringBasedStruct):
     template_dtype = np.dtype([
-        ('mean', 'f4', (3, 1)),       # x0, y0, z0
+        ('mean', 'f4', (3, 1)),  # x0, y0, z0
         ('sigma', 'f4'),
         ('m_L', 'f4', (4, 4)),
```
oesteban:
```suggestion
        ('m_L', 'f8', (4, 4)),
```
I believe that using double precision will make the conversion test more accurate. We should migrate all affines to double precision.

mgxd: ha, this was the first thing I thought of and tried with the earlier files, but there was still a large difference. will give it a try with these new transforms

oesteban: we should use double precision for all affines, anyways.

nitransforms/io/lta.py
```diff
@@ -190,7 +190,10 @@ def to_string(self):
     def from_string(klass, string):
         lta = klass()
         sa = lta.structarr
-        lines = string.splitlines()
+        lines = [l.strip() for l in string.splitlines()
+                 if l.strip() and not l.strip().startswith('#')]
+        if not lines or not lines[0].startswith('type'):
+            raise TransformFileError("Invalid LTA format")
```
oesteban: The last nitpick: can we add a test case for this?

PR #39, opened 2019-10-29
ENH: Refactor of AFNI and FSL i/o with StringStructs

oesteban:


nitransforms/io/itk.py
```diff
@@ -196,14 +184,25 @@ def to_string(self):
         strings = []
         for i, xfm in enumerate(self.xforms):
             xfm.structarr['index'] = i
-            strings.append(xfm.to_string())
+            strings.append(xfm.to_string(banner=(i == 0)))
```
mgxd: this method is missing the `banner` parameter

nitransforms/io/itk.py
```diff
@@ -226,11 +225,12 @@ def from_string(cls, string):
         lines = [l.strip() for l in string.splitlines()
                  if l.strip()]
 
-        if lines[0][0] != '#' or 'Insight Transform File V1.0' not in lines[0]:
-            raise ValueError('Unknown Insight Transform File format.')
+        if not lines or not lines[0].startswith('#') or \
+           'Insight Transform File V1.0' not in lines[0]:
```
mgxd: indentation is too close to following line - i've recently become a big fan of writing multiconditional `if`s in the form of 
```python
if (
   condition1
   and condition2
   ...
):

File: nitransforms/io/lta.py [hunk 3]
```diff
@@ -120,14 +177,15 @@ def to_string(self):
             ('{:18.15e} ' * 4).format(*sa['m_L'][2]),
             ('{:18.15e} ' * 4).format(*sa['m_L'][3]),
             'src volume info',
-            self['src'].to_string(),
+            '%s' % self['src'],
```
mgxd: i think this should work
```suggestion
            str(self['src']),
```

nitransforms/io/lta.py
```diff
@@ -120,14 +177,15 @@ def to_string(self):
...
             'dst volume info',
-            self['dst'].to_string(),
+            '%s' % self['dst'],
```
mgxd:
```suggestion
            str(self['dst']),
```

nitransforms/io/itk.py
oesteban: what do you mean? Isn't it defined on line 72?

nitransforms/io/itk.py
oesteban: thanks, I like it too! but I prefer to have the logical operators at the end of the line

nitransforms/io/itk.py
mgxd: ah, discard that - sorry for the noise

nitransforms/io/itk.py
```diff
@@ -226,11 +233,15 @@ def from_string(cls, string):
         lines = [l.strip() for l in string.splitlines()
                  if l.strip()]
 
-        if lines[0][0] != '#' or 'Insight Transform File V1.0' not in lines[0]:
-            raise ValueError('Unknown Insight Transform File format.')
+        if (
+            not lines or
+            not lines[0].startswith('#') or
+           'Insight Transform File V1.0' not in lines[0]
+           ):
```
mgxd: tiny nitpick
```suggestion
       ):
```

oesteban: flake8 doesn't like that...

PR #43, opened 2019-10-31
ENH: Base implementation of transforms chains (composition)

oesteban:
Setting up the infrastructure to allow the `+` operator, indexed transforms, etc.

This will also set up a base for the storage of transform chains.

This PR completes the requirements promised in the preprint.

### DIFF
--- a/nitransforms/base.py
+++ b/nitransforms/base.py
@@ -8,16 +8,20 @@
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Common interface for transforms."""
 from pathlib import Path
+from collections.abc import Iterable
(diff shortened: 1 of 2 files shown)


nitransforms/base.py
```diff
@@ -294,6 +317,127 @@ def _to_hdf5(self, x5_root):
...
+
+    def __getitem__(self, i):
+        """
+        Enable indexed access of transform chains.
+
+        Example
+        -------
+        >>> T1 = TransformBase()
+        >>> chain = T1 + TransformBase()
+        >>> chain[0] == T1
```
mgxd: to show its the same object
```suggestion
        >>> chain[0] is T1
```

PR #55, opened 2019-12-03
ENH: command line interface

### LINKED ISSUE #48: Command line utility
Create a script to apply transforms via the command line.

mgxd:
Closes #48 

Adds `nt` executable for command line usage of nitransforms.

Currently only supports `nt apply`, which applies a linear or nonlinear transform to an image.

### DIFF
--- a/nitransforms/cli.py
+++ b/nitransforms/cli.py
@@ -1,30 +1,49 @@
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
 from textwrap import dedent
-
-import nibabel as nb
+import sys
 
 from .linear import load as linload
 from .nonlinear import load as nlinload
 
-def cli_apply(**kwargs):
+
+def cli_apply(pargs):
     """
     Apply a transformation to an image, resampling on the reference
 
-    Usage:
+    Sample usage:
 
-        $ nt apply moving.nii.gz xform.fsl --reference reference.nii.gz
+        $ nt apply xform.fsl moving.nii.gz --ref reference.nii.gz
+
+        $ nt apply warp.nii.gz moving.nii.gz --fmt afni --nonlinear
+
     """
-    try:
-        xfm = linload(xform)
-    except:
-        xfm = nlinload(xform)
+    fmt = pargs.fmt or pargs.transform.split('.')[-1]
+    if fmt == 'tfm':
+        fmt = 'itk'
+    elif fmt == 'lta':
+        fmt = 'fs'
 
-    xfm.apply(moving, **kwargs)
+    if fmt not in ('fs', 'itk', 'fsl', 'afni'):
+        raise RuntimeError(
+            "Cannot determine transformation format, manually set format with the `--fmt` flag"
+        )
+
+    if pargs.nonlinear:
+        xfm = nlinload(pargs.transform, fmt=fmt, reference=pargs.ref)
+    else:
+        xfm = linload(pargs.transform, fmt=fmt)
+
+    xfm.apply(
+        pargs.moving,
+        order=pargs.order,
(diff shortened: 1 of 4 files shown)


nitransforms/cli.py
```diff
@@ -0,0 +1,122 @@
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
```
oesteban: click is pretty lightweight - @effigies do you think would be some resistance in nibabel to migrate `nib-ls` and other command line scripts I might not be aware of?

nitransforms/cli.py
```diff
@@ -0,0 +1,122 @@
...
+from textwrap import dedent
+import sys
+
+from .linear import load as linload
+from .nonlinear import load as nlinload
+
+
+def cli_apply(pargs):
```
oesteban: Could you add tests for this?. I would probably mock/monkeypatch xfm.apply to just check the inputs are what you expect.

nitransforms/cli.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Apply a transformation to an image, resampling on the reference
```
oesteban: Silly nitpick:
```suggestion
    Apply a transformation to an image, resampling on the reference.
```

nitransforms/cli.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Sample usage:
+
+        $ nt apply xform.fsl moving.nii.gz --ref reference.nii.gz
+
+        $ nt apply warp.nii.gz moving.nii.gz --fmt afni --nonlinear
+
+    """
+    fmt = pargs.fmt or pargs.transform.split('.')[-1]
+    if fmt == 'tfm':
```
oesteban:
```suggestion
   if fmt in ('tfm', 'mat', 'h5'):
```

nitransforms/cli.py
```diff
@@ -0,0 +1,122 @@
...
+        fmt = 'itk'
+    elif fmt == 'lta':
+        fmt = 'fs'
+
+    if fmt not in ('fs', 'itk', 'fsl', 'afni'):
```
oesteban:
```suggestion
    if fmt not in ('fs', 'itk', 'fsl', 'afni', 'x5'):
```

nitransforms/cli.py
```diff
@@ -0,0 +1,122 @@
...
+        raise RuntimeError(
```
oesteban: Perhaps ValueError?

```suggestion
        raise ValueError(
```

nitransforms/cli.py
```diff
@@ -0,0 +1,122 @@
...
+        pargs.moving,
+        order=pargs.order,
+        mode=pargs.mode,
+        cval=pargs.cval,
+        prefilter=pargs.prefilter
+    )
+
+
+def get_parser():
+    desc = dedent("""
```
oesteban: Does not argparse compose a good description for commands?

nitransforms/cli.py
effigies: In general, nibabel has tried to avoid mandatory dependencies besides `numpy`. You might get some pushback on click, but might not, given it doesn't have any dependencies of its own.

nitransforms/cli.py
mgxd: i find it's nice to have a little blurb describing the subparsers at the top level (argparse doesn't do this)

nitransforms/cli.py
mgxd: i've made the `main()` allow supplemental args to be passed in to `argparse.ArgumentParser.parse_args()` - hopefully that is sufficient for testing purposes

PR #59, opened 2020-02-18
ENH: Support for transforms mappings (e.g., head-motion correction)

### LINKED ISSUE #46: Support for TransformMap
This would be a list of Affines, for instance, to correct for head motion. In this case, the TransformMap.map(x) method (with N transforms in the map) will generate N mappings of x.

The difference with concatenation is that transforms are not sequential in the map, they are applied in parallel.

oesteban:
Implements two types of transforms mappings:

1. a general one for any internal transform, and
2. an optimized mapping for lineart transforms (head-motion correction).

Since scipy's interpn only accepts linear and nearest interpolations,
these mappings will not support for simultaneous slice-timing correction
(STC) for the time being.

The addition of 4D tensor B-Spline basis for interpolation would allow
for simultaneous STC, HMC and SDC of functional time-series.
Not sure we should keep supporting Lanczos interpolation, very much less
if we want to do all three corrections at the same time.

Closes #46


nitransforms/linear.py
```diff
@@ -187,8 +187,270 @@ def from_filename(cls, filename, fmt='X5',
...
+            fslobj = io.fsl.FSLLinearTransformArray.from_ras(
+                self.matrix, moving=moving, reference=self.reference
+            )
+            fslobj.to_filename(filename)
+            return filename
+
+        if fmt.lower() in ('fs', 'lta'):
+            # xform info
+            # to make LTA file format
+            lta = io.LinearTransformArray()
```
mgxd: should we rename this to `FSLinearTransformArray` to be uniform with the other LTAs?

nitransforms/linear.py
```diff
@@ -187,8 +187,270 @@ def from_filename(cls, filename, fmt='X5',
...
+            lta['type'] = 1  # RAS2RAS
+            for m in self.matrix:
+                lt = io.LinearTransform()
```
mgxd: and this to `FSLinearTransform`?

nitransforms/tests/test_linear.py
```diff
@@ -31,6 +31,64 @@
...
 
 
+@pytest.mark.parametrize('matrix', [
+    [0.0],
+    np.ones((3, 3, 3)),
+    np.ones((3, 4)),
+])
+def test_linear_typeerrors1(matrix):
+    """Exercise errors in Affine creation."""
+    with pytest.raises(TypeError):
```
mgxd:
```suggestion
    with pytest.raises(ValueError):
```

nitransforms/tests/test_linear.py
```diff
@@ -31,6 +31,64 @@
...
+        ntl.Affine(matrix)
+
+
+def test_linear_typeerrors2(data_path):
+    """Exercise errors in Affine creation."""
+    with pytest.raises(TypeError):
```
mgxd:
```suggestion
    with pytest.raises(ValueError):
```

nitransforms/linear.py
```diff
@@ -35,34 +42,29 @@ def __init__(self, matrix=None, reference=None):
...
-        ), 'affine matrix is not square'
+        super().__init__(reference=reference)
+        self._matrix = np.eye(4)
 
-        if reference:
-            self.reference = reference
+        if matrix is not None:
+            matrix = np.array(matrix)
+            if matrix.ndim != 2:
+                raise TypeError('Affine should be 2D.')
```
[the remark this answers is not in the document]
oesteban: Since this matrix is the core formal element and the problem is not whether the values in the matrix make sense or are from a valid finding of inputs, this seemed to me more like a type error. Let me know if this argument is convincing

nitransforms/linear.py
oesteban: I think we can keep lta. But, if we were to change it, let's in a different pr

nitransforms/linear.py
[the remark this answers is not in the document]
mgxd: At least to me, it doesn't make sense to throw a `TypeError` if the matrix is an array, but it does make sense to throw a `ValueError` if the array dimensions aren't what is expected. But your call at end, nbd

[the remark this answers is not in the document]
oesteban: Okay, this is my last attempt to show what I think is a `ValueError` in contrast to a `TypeError`: https://github.com/poldracklab/nitransforms/pull/59/files#diff-ca10ddac142c49ab69b356fb51a6a075R69-R70

Here, all the formal checks (i.e., shape) on the validity of the matrix pass, but the values in it are outside of the possible inputs domain.

The two type errors above could go away if we modify this object to, for instance, take a (3, 4) matrix (à la AFNI), or accept a mat+vec representation (allow the argument matrix to be (3, 3) and have an extra (3, ) vector argument for the translation).

The ValueError,  though, will always be there.

This is completely a conceptual and possibly arbitrary argument - if you are still unconvinced I'll assume I'm wrong (that's the most likely possibility in that case because you are an English speaker and I'm not) and switch all to `ValueError`s.

nitransforms/linear.py
```diff
@@ -187,8 +187,270 @@ def from_filename(cls, filename, fmt='X5',
 
         struct = _factory.from_filename(filename)
         matrix = struct.to_ras(reference=reference, moving=moving)
-
+        if cls == Affine:
+            if np.shape(matrix)[0] != 1:
+                raise TypeError(
```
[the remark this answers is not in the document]
oesteban: Hmmm I forgot about this one. This is indeed a ValueError under my own definitions.

PR #60, opened 2020-02-19
MAINT: Better config on setup.py (binary operator starting line)

oesteban:


setup.cfg
```diff
@@ -33,5 +33,22 @@ console_scripts =
     nt = nitransforms.cli:main
 
 [flake8]
```
mgxd: since we plan on porting this over to nibabel, let's revert the style checks to match theirs

nitransforms/tests/test_nonlinear.py
```diff
@@ -11,7 +11,7 @@
 from ..io.base import TransformFileError
 from ..nonlinear import DisplacementsFieldTransform, load as nlload
 from ..io.itk import ITKDisplacementsField
-from ..io.afni import AFNIDisplacementsField
+# from ..io.afni import AFNIDisplacementsField
```
mgxd:
```suggestion
```

setup.cfg
oesteban: ugh right. We might want to push for more strict flake8 checks. @effigies ?

oesteban: Let's go-ahead - good syntax should not be a problem when integrating into nibabel.

effigies: Ah, sorry. I saw this on my notifications and then it disappeared. Yeah, we can improve things in nibabel, but being more strict here can't cause problems.

PR #63, opened 2020-02-25
FIX: Amend some error types

oesteban:
Trying to normalize a little between `TypeError`s and `ValueError`s.


oesteban:
Let's check on all error types when merging into nibabel (#110).

Issue #64, opened 2020-03-18
Issue reading an LTA generated by fMRIPrep


oesteban:
[This LTA file](https://github.com/poldracklab/nitransforms/files/4347227/T1_robustreg.txt) ([generated by this node](https://github.com/poldracklab/smriprep/blob/b5b74ad2dbd54284f36ca843829ec38014637c2a/smriprep/workflows/outputs.py#L238)) fails to load:
```Python
>>> import nitransforms as nit
>>> xfm = nit.linear.load('T1_robustreg.lta', fmt='lta')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/oesteban/workspace/nitransforms/nitransforms/linear.py", line 455, in load
    filename, fmt=fmt, reference=reference, moving=moving
  File "/home/oesteban/workspace/nitransforms/nitransforms/linear.py", line 198, in from_filename
    return cls(matrix, reference=reference)
  File "/home/oesteban/workspace/nitransforms/nitransforms/linear.py", line 236, in __init__
    for xfm in transforms
  File "/home/oesteban/workspace/nitransforms/nitransforms/linear.py", line 236, in <listcomp>
    for xfm in transforms
  File "/home/oesteban/workspace/nitransforms/nitransforms/linear.py", line 70, in __init__
    raise ValueError("Matrix does not represent a valid transform.")
ValueError: Matrix does not represent a valid transform.
```

oesteban:
BTW - that transform, as per fMRIPrep, translates to ITK as [this file](https://github.com/poldracklab/nitransforms/files/4347270/sub-01_from-fsnative_to-T1w_mode-image_xfm.txt) - we should add a test checking idempotence (i.e., lta <-> itk <-> lta).

oesteban:
It seems this is still failing at "/home/oesteban/workspace/nitransforms/nitransforms/linear.py", line 70.

PR #66, opened 2020-03-19
ENH: Add ITK-LTA conversion test

oesteban:
## Summary

This PR adds a few tests, and also addresses an uncovered cause of #64.

## This PR assumes that the matrix found in LTA transforms is given in ``mode-points`` instead of ``mode-images`` (in BIDS-Derivatives terms).
Further investigation of this would be advised, this is just the empirical result of using fMRIPrep-generated LTAs (and ITK conversions) for the new tests. 

## TODO
I have added an FSL affine generated from an LTA (bbregister) - Since we don't have references for it, I couldn't add a test this time.


oesteban:
@effigies @mgxd I have edited the PR's description and some of the commits, to cover more comprehensively that I believe the LTA matrices are ``mode-points``.

nitransforms/tests/data/regressions/robust_register.lta
```diff
@@ -1,30 +1,28 @@
-# transform file /home/oesteban/tmp/fmriprep-ds005/fprep-work/fmriprep_wf/single_subject_01_wf/anat_preproc_wf/surface_recon_wf/fsnative2t1w_xfm/T1_robustreg.lta
-# created by oesteban on Sat Mar 14 19:28:37 2020
-
```
effigies: Does this new file fail prior to #65?

oesteban: Yes, I have updated both because I had messed up when filing #64. The attached ITK file I believe is the inverse. To be sure, I searched again for both (LTA, ITK) from the robust register node and subsequent nodes and replaced them.

PR #74, opened 2020-03-19
ENH: Add FS transform regression

mgxd:
Tests are passing locally, but a bit iffy on the results.


effigies:
Looks like we're not passing the freesurfer license into the Docker container?

mgxd:
> there must be some discrepancy with your local installation

WDYM?

oesteban:
> WDYM

The license - you had it only locally :D. Merging...

PR #77, opened 2020-03-20
MAINT: Drop Python 3.5

oesteban:


mgxd:
Are we still planning on merging this with nibabel? This pushes ahead of it

oesteban:
I've thought about that. By the time this can be merged into nibabel, I bet the support will be for +3.6 only. Otherwise, I think it is worth making the development more agile and figure out when the time comes to join both projects.

PR #79, opened 2020-03-25
ENH: Read (and apply) ITK/ANTs' composite HDF5 transforms

### LINKED ISSUE #4: Read in ITK's h5 transforms

oesteban:
Uses ``h5py`` to browse the transforms group, and concatenates them in a list
ready for nitransforms' wrapping into a ``TransformsChain``.

The writer is not included in this PR.

The implementation takes into account the misspelling of "Tranforms" present
in transforms generated with older versions of ITK.

This implementation only accepts the composite transforms generated by ANTs.

Closes #4


nitransforms/tests/test_manip.py
```diff
@@ -0,0 +1,54 @@
...
+
+    xfm = _load(xfm_fname)
+
+    assert len(xfm) == 2
+
+    ref_fname = tmp_path / "reference.nii.gz"
+    nb.Nifti1Image(
+        np.zeros(xfm.reference.shape, dtype='uint16'),
+        xfm.reference.affine,
+        None
```
mgxd: this is optional so no need to include
```suggestion
```

PR #102, opened 2021-03-24
FIX: Accept double precision when reading ANTs' ``.mat`` affines

oesteban:


nitransforms/io/itk.py
```diff
@@ -129,10 +129,15 @@ def from_matlab_dict(cls, mdict, index=0):
         tf = cls()
         sa = tf.structarr
 
+        affine = mdict.get(
+            "AffineTransform_double_3_3",
+            mdict.get("AffineTransform_float_3_3")
+        )
```
effigies, line 135: Is one or the other guaranteed to exist at this point? Or should we check `if affine is None` after?

effigies (approved):
Overall looks reasonable. Appears that this implies that we will always output double affines. Do all tools accept those, or is there a tradeoff here?

nitransforms/io/itk.py
oesteban: Theoretically, ITK allows you any combination of "Translation", "Euler", "Versors", "Rigid", "Similarity" and "Affine" with the data type string and the dimensions numbers.

However, I don't think ANTs produces anything other than Affine.

I'll raise though, to keep the user better informed.

oesteban:
It is a tradeoff, after all this is a 16 parameter matrix so float64 is not a big deal. However, I think I recall @hjmjohnson saying that it is critical that affines preserve the precision on some occasions at our transforms meeting in Boston two years ago.

oesteban:
> I meant a tradeoff in the sense of losing compatibility with other tools

No, perhaps the opposite. Also, this format is kind of deprecated for ANTS - at some point they will move completely to HDF5. If they haven't done so yet it is probably because of the low priority of the move.

Issue #45, opened 2019-10-31
Implement AFNI affines on oblique datasets


oesteban:
Our framework chokes in this use case. Prior knowledge on the topic is found here: https://github.com/afni/afni/issues/116

PR #117, opened 2021-07-02
FIX: Implement AFNI's deoblique operations

oesteban:
Finalize the interpretation in I/O of AFNI's linear matrices when reference
or moving data are oblique w.r.t. the canonical axes.

Resolves: #45.


oesteban:
Okay, this is as far as I can go for the moment (see afni/afni#353). I'm going to focus on other things while I wait for more feedback from AFNI folks. 

/cc @effigies @matthew-brett

docs/notebooks/01_preparing_images.ipynb
```diff
@@ -745,9 +757,11 @@
     }
    ],
    "source": [
-    "!3dAllineate -base someones_anatomy_rot.nii.gz -input someones_anatomy_rot.nii.gz -1Dmatrix_apply M.afni -prefix moved-afni.nii.gz -final NN\n",
-    "nb.load('moved-afni.nii.gz').orthoview()\n",
-    "moved.orthoview()"
+    "afnixfm = nt.io.afni.AFNILinearTransform.from_filename('M.afni')\n",
+    "xfm = nt.linear.Affine(afnixfm.to_ras(), reference=deobnii)\n",
+    "nt_with_afni = xfm.apply(plumb, order=0)\n",
```
effigies, line 762: plumb is not defined.

effigies (commented):
The notebooks don't run all the way through. Was hoping to use them to guide my review.

oesteban:
Ouch, yes, I think they are not necessary anymore but I agree they are
useful to understand some of the tests. Will try to rework them a little
for your review.

effigies:
If they're not necessary, that's fine. Just assumed that with their changes they were intended to curate the situation.

oesteban:
I was thinking about removing the notebook 02 about AFNI, and rewriting 01
to focus on stuff from the io submodule.

oesteban:
Okay, I think now the notebooks are in much better shape. LMKWYT.

oesteban:
I would imagine you are really up to your eyeballs @effigies - should I wait for review or can I go ahead and maybe you can catch up later when the integration with nibabel starts?

effigies:
Was just starting to have a quick look before our call, and planned to finish up after. I'm okay if you want to go ahead and merge, but I did want to suggest that it might make sense to think of affines generally as `T @ R_O @ R_C @ Z @ S` where `R_C` is the cardinal rotation and `R_O` is an additional oblique rotation `<90deg`. I haven't thought through the details, so it might be too cumbersome, but it seems like this conceptualization could allow us to write code that is more similar across tools and make the specific deviations of AFNI clear in the algebra, rather than via special functions.

nitransforms/io/afni.py
```diff
@@ -36,33 +38,17 @@ def to_string(self, banner=True):
     @classmethod
     def from_ras(cls, ras, moving=None, reference=None):
         """Create an AFNI affine from a nitransform's RAS+ matrix."""
```
effigies, line 40: Something like the following would help me understand the adjustment to the RAS:

```suggestion
        """Create an AFNI affine from a nitransform's RAS+ matrix.

        AFNI implicitly de-obliques image affine matrices before applying transforms, so
        for consistency we update the transform to account for the obliquity of the images.
        
        .. testsetup:
            >>> import pytest
            >>> pytest.skip()

        >>> moving.affine == ras @ reference.affine

        We can decompose the affines into oblique and de-obliqued components:

        >>> moving.affine == m_obl @ m_deobl
        >>> reference.affine == r_obl @ r_deobl
        
        To generate an equivalent AFNI transform, we need an effective transform (``e_ras``):
        
        >>> m_obl @ m_deobl == ras @ r_obl @ r_deobl
        >>> m_deobl == inv(m_obl) @ ras @ r_obl @ r_deobl
        
        Hence,
        
        >>> m_deobl == e_ras @ r_deobl
        >>> e_ras == inv(m_obl) @ ras @ r_obl
        """
```

I wrote it in Python, but LaTeX would be better. I just didn't want to try to do the syntax.

nitransforms/io/afni.py
```diff
@@ -36,33 +38,17 @@ def to_string(self, banner=True):
...
-            M2 = moving.affine
-            A2 = shape_zoom_affine(
-                moving.shape, voxel_sizes(M2), x_flip=True, y_flip=True
-            )
-            post = A2.dot(np.linalg.inv(M2))
+            ras = _cardinal_rotation(moving.affine, True) @ ras
 
-        # swapaxes is necessary, as axis 0 encodes series of transforms
-        parameters = np.swapaxes(post @ ras @ pre, 0, 1)
+        parameters = np.swapaxes(LPS @ ras @ LPS, 0, 1)
```
effigies, line 51:
```suggestion
        # AFNI represents affine transformations as LPS-to-LPS
        parameters = np.swapaxes(LPS @ ras @ LPS, 0, 1)
```

effigies (approved):
Okay, the code makes sense once I worked through the algebra, which I coded up as a docstring so others don't have to do it again. I read the tests with an eye to get the gist, but did not verify that all comments correctly described what was being done. This looks reasonable to me.

Issue #123, opened 2021-07-14
JOSS review - software 


robbisg:
Hi,

this issue is related to JOSS submission (https://github.com/openjournals/joss-reviews/issues/3459) and focuses more on the software side.

I think that this package can be very useful to easily integrate different formats and algorithms into more complicated pipelines, but also with the adoption of BIDS specification, so thanks for sharing this package with the community.

I noticed that you added the documentation website with the API of the tools, you have a set of test with a very high coverage and automated CI, and it's great, since it takes lot of time to code these parts! 

Some points needs to be added in my opinion (but also from JOSS guidelines :) ):
* __examples__: from a user point of view, as you know, this will help users to understand and use (correctly) your package.
* __installation guidelines__: I suppose that the good old `pip install` or `python setup.py install` are used to install the package, but this should be mentioned in the repo README with dependencies.
* __contribution guidelines__: this is a minor edit since you planned to merge this package into `nibabel` which have this section, but a couple of lines in the docs / README are needed until this package will be independent and for JOSS, as well. 

I saw that some of these concerns have been answered in #118 .

I have an extra question, that is more a curiosity than a issue related to the tool: do you think it would be useful to include in your package some atlas transformations utilities (e.g. `to_talairach`, `to_mni`) or is out of the scope of it?

Cheers,
Roberto

PS: I saw while I was writing that some of these issues have been addressed by #122

robbisg:
## Paper review

I was opening another issue related to paper edits, but I saw that in #121 you just edited the paper and addressed lot of stuff there.

I have an extra suggestion: a comparison with other tools that are already available is needed for JOSS, and I think can help to understand the importance of your tool. So you can add a sentence if yours is the first the tool that address the conversion of transformation file formats, or add a short comparison with other tools.

oesteban:
> very high coverage and automated CI, and it's great, since it takes lot of time to code these parts!

This was part of the vision - the reliability our implementation requires is only possible to achieve under test-driven development approaches.

> * **examples**: from a user point of view, as you know, this will help users to understand and use (correctly) your package.

Thanks - yes this is a great suggestion. We already have some notebooks, I'll see how to integrate them into the docs ASAP. We can also add a link to https://www.nipreps.org/nipreps-book/tutorial/registration.html, where the API is demonstrated in a real scenario. Finally, we have a binder link - the container is pretty heavy so it fails often, but it is worth adding somewhere and optimize the container down the line.

[...]

PR #127, opened 2021-07-19
DOC: Add CONTRIBUTING.md file pointing at NiBabel

oesteban:
Linked: #123


mgxd:
@oesteban is this not visible from the docs?

oesteban:
Yes, good catch. We should move it somewhere it makes sense

Issue #125, opened 2021-07-15
BUG: Errors reading LTAs, especially interpreting VOX2VOX type


oesteban:
For instance, when reading:

```
# transform file /work/fmriprep_wf/single_subject_10316_wf/func_preproc_task_bart_wf/bold_reg_wf/bbreg_wf/bbregister/uni_xform_masked_bbreg_sub-10316.lta
# created by unknown on Wed Jul 14 10:10:37 2021

type      = 0 # LINEAR_VOX_TO_VOX
nxforms   = 1
mean      = 0.0000 0.0000 0.0000
sigma     = 10000.0000
1 4 4
2.999451637268066e+00 1.401854865252972e-02 -7.407368719577789e-02 3.591539764404297e+01 
-2.879538573324680e-02 -2.145972728729248e+00 -2.794905662536621e+00 2.451114654541016e+02 
-4.953517019748688e-02 2.096330165863037e+00 -2.860595941543579e+00 9.669815826416016e+01 
0.000000000000000e+00 0.000000000000000e+00 0.000000000000000e+00 1.000000000000000e+00 
src volume info
valid = 1  # volume info valid
filename = /work/fmriprep_wf/single_subject_10316_wf/func_preproc_task_bart_wf/bold_reg_wf/bbreg_wf/bbregister/uni_xform_masked.nii.gz
volume = 64 64 34
voxelsize = 3.000000000000000e+00 3.000000000000000e+00 4.000000000000000e+00
xras   = -1.000000000000000e+00 -0.000000000000000e+00 0.000000000000000e+00
yras   = 0.000000000000000e+00 6.858183145523071e-01 7.277727723121643e-01
zras   = -0.000000000000000e+00 -7.277727723121643e-01 6.858183741569519e-01
cras   = -4.697326660156250e+00 -9.175056457519531e+00 1.141981506347656e+01
dst volume info
valid = 1  # volume info valid
filename = /freesurfer/sub-10316/mri/orig.mgz
volume = 256 256 256
voxelsize = 1.000000000000000e+00 1.000000000000000e+00 1.000000000000000e+00
xras   = -1.000000000000000e+00 0.000000000000000e+00 0.000000000000000e+00
yras   = 0.000000000000000e+00 0.000000000000000e+00 -1.000000000000000e+00
zras   = 0.000000000000000e+00 1.000000000000000e+00 0.000000000000000e+00
[...]

PR #129, opened 2021-07-19
FIX: Misinterpretation of voxel ordering in LTAs

oesteban:
The structarray was used directly and the extra axis actually changed the order of operations when the direction cosines were scaled by the voxel sizes.

A new test case has been added for which this error was apparent. This bug caused nipreps/fmriprep#2307, nipreps/fmriprep#2393, and nipreps/fmriprep#2410. nipreps/fmriprep#2444 just works around the problem by using ``lta_convert`` instead of *NiTransforms*. The ``lta_convert`` tool can be now dropped.

Resolves: #125


nitransforms/io/lta.py
```diff
@@ -39,10 +39,10 @@ def as_affine(self):
...
         sa = self.structarr
-        A = np.hstack((sa["xras"], sa["yras"], sa["zras"])) * sa["voxelsize"]
-        b = sa["cras"] - A.dot(sa["volume"]) / 2
-        affine[:3, :3] = A
-        affine[:3, [3]] = b
+        affine[:3, :3] = (
+            np.hstack((sa["xras"], sa["yras"], sa["zras"])) * sa["voxelsize"].flatten()
+        )
+        affine[:3, [3]] = sa["cras"] - affine[:3, :3] @ sa["volume"] / 2
         return affine
```
effigies, line 46: I feel like this loses clarity. Right now we're decomposing `Mx = Ax + b`, and showing how A and b interact.

Also, if I'm understanding you correctly, the problem is that we're applying `voxelsize` row-wise when it should be column-wise, which suggests the correct fix is:

```Diff
-        A = np.hstack((sa["xras"], sa["yras"], sa["zras"])) * sa["voxelsize"]
+        A = np.hstack((sa["xras"], sa["yras"], sa["zras"])) @ sa["voxelsize"]
```

The reason that the header fields are shaped as they are is to make the computations straightforward linear algebra. If it's not useful, we can just flatten them into shape `(3,)` instead of `(3, 1)`.

Finally, however we do this, could we add a test with a geometry that shows the old formulation was wrong and the new correct? I get the regression tests are checking this, but something readable, with, e.g. voxel sizes like `(2, 3, 4)` and cosines that are ones/zeros would be nice.

nitransforms/tests/test_io.py
```diff
@@ -37,6 +38,22 @@ def test_VolumeGeometry(tmpdir, get_testdata):
...
+-3.0000   0.0000  -0.0000    91.3027
+-0.0000   2.0575  -2.9111   -25.5251
+ 0.0000   2.1833   2.7433  -105.0820
+ 0.0000   0.0000   0.0000     1.0000"""))
+
+    lta_text = "\n".join(
+        (data_path / "bold-to-t1w.lta").read_text().splitlines()[13:21]
+    )
+    r2r = VG.from_string(lta_text)
+    assert np.allclose(r2r.as_affine(), oracle, rtol=1e-4)
```
effigies, line 54: AFAICT this test doesn't fail before this PR. What exactly is the edge case we're fixing here?

oesteban: Yes, this failed before this PR.

oesteban: The test breaks current master without the patch proposed -> https://github.com/poldracklab/nitransforms/compare/fix/vg-lta-test

nitransforms/io/lta.py
oesteban:
```Python
A = np.hstack((sa["xras"], sa["yras"], sa["zras"])) * sa["voxelsize"]
```
If the dimensions of `sa["voxelsize"]` are (3, 1) then this is row-wise product. Conversely, if the dimension is (3, ) then this is col-wise product. We want column wise product and before this patch we were doing row wise product.

Dot product doesn't apply here, as it rotates the voxelsize vector through the affine.

oesteban:
Good to go? Do you feel the comments have been successfully addressed?

effigies:
Sorry, just getting back to this this morning. Give me one hour.

nitransforms/tests/test_io.py
effigies, line 54: Here's the kind of test I was thinking:

```suggestion
def test_volume_group_voxel_ordering():
    vg = VG.from_string("""\
valid = 1  # volume info valid
filename = no_file
volume = 5 6 7
voxelsize = 2 3 4
xras   = -1 0 0
yras   = 0 0 1
zras   = 0 -1 0
cras   = 0 0 0""")
    aff = vg.as_affine()
    assert np.allclose(vg["voxelsize"], [2, 3, 4])
    assert np.allclose(nb.affines.voxel_sizes(aff), [2, 3, 4])
    assert nb.aff2axcodes(aff) == ("L", "S", "P")
```

The property we're testing is that the zooms are correctly applied, so I'd rather be explicit than have it vague what qualities about the problematic geometry have been fixed.

nitransforms/io/lta.py
```diff
@@ -24,26 +24,23 @@ class VolumeGeometry(StringBasedStruct):
...
-        affine = np.eye(4)
         sa = self.structarr
-        A = np.hstack((sa["xras"], sa["yras"], sa["zras"])) * sa["voxelsize"]
-        b = sa["cras"] - A.dot(sa["volume"]) / 2
-        affine[:3, :3] = A
-        affine[:3, [3]] = b
-        return affine
+        R = np.hstack((sa["xras"], sa["yras"], sa["zras"])) * sa["voxelsize"]
+        offset = sa["cras"] - R @ sa["volume"] / 2
+        return from_matvec(R, offset)
```
effigies, line 43: We should use a consistent notation here and in L74-80. I prefer A/b because this is a common representation of an augmented matrix (see [Affine transformation - Representation](https://en.wikipedia.org/wiki/Affine_transformation#Representation)).

In any event, I wouldn't use R, because that means rotation, and this matrix has scale and rotation. (Not sure if you can sneak shear into the MGH format in practice.)

```suggestion
        A = np.hstack((sa["xras"], sa["yras"], sa["zras"])) * sa["voxelsize"]
        b = sa["cras"] - A @ sa["volume"] / 2
        return from_matvec(A, b)
```

nitransforms/tests/test_io.py
oesteban: Sure, having an extra test and it being clear can only help.

However, I would keep the test I added for two reasons:
- First, because this is one example we know at some point of history failed.
- Second, because the test I wrote is dual to some test freesurfer folks could add to their repo (or, better said, I bet they already are doing something like that). Because `oracle` comes from executing `mri_info` (I saved the part of actually running it on every test), then we are able to at least ensure that our volume geometry interpretation is the same of freesurfer 6.0.1.

effigies: Sure, we can keep your existing test.

nitransforms/io/lta.py
```diff
@@ -72,15 +69,15 @@ def from_image(klass, img):
         volgeom = klass()
         sa = volgeom.structarr
         sa["valid"] = 1
-        sa["volume"][:, 0] = img.shape[:3]  # Assumes xyzt-ordered image
-        sa["voxelsize"][:, 0] = voxel_sizes(img.affine)[:3]
+        sa["volume"] = img.shape[:3]  # Assumes xyzt-ordered image
+        sa["voxelsize"] = voxel_sizes(img.affine)[:3]
         A = img.affine[:3, :3]
-        b = img.affine[:3, [3]]
+        b = img.affine[:3, [3]].flatten()
```
effigies, line 75:
```suggestion
        b = img.affine[:3, 3]
```

effigies (approved):
Looks good. One tiny suggestion to avoid a call to `flatten()`.

Issue #106, opened 2021-06-16
Bring B-Spline interpolation code from SDCFlows


oesteban:
[nipreps/sdcflows](https://github.com/nipreps/sdcflows) has been a wonderful testbed for the interpolation of non-gridded data on a B-Spline basis field. It is probably time to bring it here.

PR #138, opened 2022-02-10
ENH: Base implementation of B-Spline transforms

oesteban:
Addressing #106.

### DIFF
--- a/nitransforms/interp/bspline.py
+++ b/nitransforms/interp/bspline.py
@@ -1,0 +1,95 @@
+# emacs: -*- mode: python-mode; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the NiBabel package for the
+#   copyright and license terms.
+#
+### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""Interpolate with 3D tensor-product B-Spline basis."""
+import numpy as np
+import nibabel as nb
+from scipy.sparse import csr_matrix, kron
+
+
+def _cubic_bspline(d, order=3):
+    """Evaluate the cubic bspline at distance d from the center."""
+    if order != 3:
+        raise NotImplementedError
+
+    return np.piecewise(
+        d,
+        [d < 1.0, d >= 1.0],
+        [
+            lambda d: (4.0 - 6.0 * d ** 2 + 3.0 * d ** 3) / 6.0,
+            lambda d: (2.0 - d) ** 3 / 6.0,
+        ],
+    )
+
+
+def grid_bspline_weights(target_nii, ctrl_nii):
+    r"""
+    Evaluate tensor-product B-Spline weights on a grid.
+
+    For each of the *N* input samples :math:`(s_1, s_2, s_3)` and *K* control
+    points or *knots* :math:`\mathbf{k} =(k_1, k_2, k_3)`, the tensor-product
+    cubic B-Spline kernel weights are calculated:
+
+    .. math::
+        \Psi^3(\mathbf{k}, \mathbf{s}) =
+        \beta^3(s_1 - k_1) \cdot \beta^3(s_2 - k_2) \cdot \beta^3(s_3 - k_3),
+        \label{eq:1}\tag{1}
+
+    where each :math:`\beta^3` represents the cubic B-Spline for one dimension.
+    The 1D B-Spline kernel implementation uses :obj:`numpy.piecewise`, and is based on the
+    closed-form given by Eq. (6) of [Unser1999]_.
+
+    By iterating over dimensions, the data samples that fall outside of the compact
+    support of the tensor-product kernel associated to each control point can be filtered
+    out and dismissed to lighten computation.
+
+    Finally, the resulting weights matrix :math:`\Psi^3(\mathbf{k}, \mathbf{s})`
+    can be easily identified in Eq. :math:`\eqref{eq:1}` and used as the design matrix
+    for approximation of data.
+
+    Parameters
+    ----------
+    target_nii :  :obj:`nibabel.spatialimages`
+        An spatial image object (typically, a :obj:`~nibabel.nifti1.Nifti1Image`)
+        embedding the target EPI image to be corrected.
+        Provides the location of the *N* samples (total number of voxels) in the space.
+    ctrl_nii : :obj:`nibabel.spatialimages`
+        An spatial image object (typically, a :obj:`~nibabel.nifti1.Nifti1Image`)
+        embedding the location of the control points of the B-Spline grid.
+        The data array should contain a total of :math:`K` knots (control points).
+
+    Returns
+    -------
+    weights : :obj:`numpy.ndarray` (:math:`K \times N`)
+        A sparse matrix of interpolating weights :math:`\Psi^3(\mathbf{k}, \mathbf{s})`
+        for the *N* voxels of the target EPI, for each of the total *K* knots.
+        This sparse matrix can be directly used as design matrix for the fitting
+        step of approximation/extrapolation.
+
+    """
+    shape = target_nii.shape[:3]
+    ctrl_sp = ctrl_nii.header.get_zooms()[:3]
+    ras2ijk = np.linalg.inv(ctrl_nii.affine)
+    origin = nb.affines.apply_affine(ras2ijk, [tuple(target_nii.affine[:3, 3])])[0]
+
+    wd = []
+    for i, (o, n, sp) in enumerate(
+        zip(origin, shape, target_nii.header.get_zooms()[:3])
+    ):
+        locations = np.arange(0, n, dtype="float16") * sp / ctrl_sp[i] + o
+        knots = np.arange(0, ctrl_nii.shape[i], dtype="float16")
+        distance = np.abs(locations[np.newaxis, ...] - knots[..., np.newaxis])
+
+        within_support = distance < 2.0
+        d_vals, d_idxs = np.unique(distance[within_support], return_inverse=True)
+        bs_w = _cubic_bspline(d_vals)
+        weights = np.zeros_like(distance, dtype="float32")
+        weights[within_support] = bs_w[d_idxs]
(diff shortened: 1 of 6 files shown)


nitransforms/interp/bspline.py
```diff
@@ -0,0 +1,95 @@
...
+    shape = target_nii.shape[:3]
+    ctrl_sp = nb.affines.voxel_sizes(ctrl_nii.affine)[:3]
+    ras2ijk = np.linalg.inv(ctrl_nii.affine)
+    origin = nb.affines.apply_affine(ras2ijk, [tuple(target_nii.affine[:3, 3])])[0]
+
+    wd = []
+    for i, (o, n, sp) in enumerate(
+        zip(origin, shape, nb.affines.voxel_sizes(target_nii.affine)[:3])
+    ):
+        locations = np.arange(0, n, dtype="float16") * sp / ctrl_sp[i] + o
```
effigies, line 84:
```suggestion
        # Locations of voxels in target image in control point image
        locations = np.arange(0, n, dtype="float16") * sp / ctrl_sp[i] + o
```

nitransforms/interp/bspline.py
```diff
@@ -0,0 +1,95 @@
...
+        A sparse matrix of interpolating weights :math:`\Psi^3(\mathbf{k}, \mathbf{s})`
+        for the *N* voxels of the target EPI, for each of the total *K* knots.
+        This sparse matrix can be directly used as design matrix for the fitting
+        step of approximation/extrapolation.
+
+    """
+    shape = target_nii.shape[:3]
+    ctrl_sp = nb.affines.voxel_sizes(ctrl_nii.affine)[:3]
+    ras2ijk = np.linalg.inv(ctrl_nii.affine)
+    origin = nb.affines.apply_affine(ras2ijk, [tuple(target_nii.affine[:3, 3])])[0]
```
effigies, line 78: Mostly my own notes, but might help future readers.

```suggestion
    # IJK index in the control point image of the first index in the target image
    origin = nb.affines.apply_affine(ras2ijk, [tuple(target_nii.affine[:3, 3])])[0]
```

nitransforms/nonlinear.py
```diff
@@ -33,13 +57,20 @@ def __init__(self, field, reference=None):
                 "the number of dimensions (%d)" % (self._field.shape[-1], ndim)
             )
 
-        self.reference = field.__class__(
-            np.zeros(self._field.shape[:-1]), field.affine, field.header
-        )
+    def __repr__(self):
+        """Beautify the python representation."""
+        return f"<{self._field.shape[:3]} field of {self._field.shape[-1]}D displacements>"
```
effigies, line 62:
```suggestion
        return f"<DisplacementFieldTransform[{self._field.shape[-1]}D] {self._field.shape[:3]}>"
```

nitransforms/nonlinear.py
```diff
@@ -22,9 +31,24 @@ class DisplacementsFieldTransform(TransformBase):
...
 
     def __init__(self, field, reference=None):
-        """Create a dense deformation field transform."""
+        """
+        Create a dense deformation field transform.
+
+        Example
+        -------
+        >>> DisplacementsFieldTransform(test_dir / "someones_displacement_field.nii.gz")
+        <(57, 67, 56) field of 3D displacements>
```
effigies, line 40: Generally expect class name to come first.

```suggestion
        <DisplacementFieldTransform[3D] (57, 67, 56)>
```

nitransforms/nonlinear.py
```diff
@@ -22,9 +31,24 @@ class DisplacementsFieldTransform(TransformBase):
...
+
+        """
         super().__init__()
-        self._field = np.asanyarray(field.dataobj)
+
+        field = _ensure_image(field)
+        self._field = np.squeeze(
+            np.asanyarray(field.dataobj) if hasattr(field, "dataobj") else field
+        )
```
effigies, line 48: No need to change anything, but I think `_ensure_image` doesn't quite capture the operation of something that will pass ndarrays through untouched.

nitransforms/nonlinear.py
```diff
@@ -22,9 +31,24 @@ class DisplacementsFieldTransform(TransformBase):
...
+        self.reference = reference or field.__class__(
+            np.zeros(self._field.shape[:-1]), field.affine, field.header
```
effigies, line 50: This will fail if `field` is an `ndarray`.

nitransforms/nonlinear.py
```diff
@@ -90,70 +118,162 @@ def from_filename(cls, filename, fmt="X5"):
...
+
+    def __init__(self, coefficients, reference=None, order=3):
+        """Create a smooth deformation field using B-Spline basis."""
+        super(BSplineFieldTransform, self).__init__()
+        self._order = order
+
+        coefficients = _ensure_image(coefficients)
+
+        self._coeffs = np.asanyarray(coefficients.dataobj)
+        self._knots = ImageGrid(four_to_three(coefficients)[0])
```
effigies, line 135: Seems wasteful to create a bunch of new images to take the first and throw it away almost immediately. ImageGrid already resets to `shape = shape[:3]` if `ndim == 4`.

```suggestion
        self._knots = ImageGrid(coefficients)
```

effigies (commented):
Note to self: pick back up at `to_field()`.

nitransforms/nonlinear.py
```diff
@@ -90,70 +118,162 @@ def from_filename(cls, filename, fmt="X5"):
...
+        reference = _ensure_image(reference)
+        _ref = self.reference if reference is None else SpatialReference.factory(reference)
+        if _ref is None:
+            raise ValueError("A reference must be defined")
+
+        ndim = self._coeffs.shape[-1]
+
+        # If locations to be interpolated are on a grid, use faster tensor-bspline calculation
+        if self._weights is None:
+            self._weights = grid_bspline_weights(_ref, self._knots)
```
effigies, line 156: Docstring says that `ctrl_nii` (`self._knots`) should be a SpatialImage, not an `ImageGrid`. It just needs `affine` and `shape`, so this is fine. Should just update the docstring.

nitransforms/interp/bspline.py
```diff
@@ -0,0 +1,95 @@
...
+        ],
+    )
+
+
+def grid_bspline_weights(target_nii, ctrl_nii):
+    r"""
+    Evaluate tensor-product B-Spline weights on a grid.
+
+    For each of the *N* input samples :math:`(s_1, s_2, s_3)` and *K* control
+    points or *knots* :math:`\mathbf{k} =(k_1, k_2, k_3)`, the tensor-product
```
effigies, line 35: This math seems off. In the equation, you have `s_i - k_i`, but nothing here constrains `N = K`.

nitransforms/nonlinear.py
```diff
@@ -90,70 +118,162 @@ def from_filename(cls, filename, fmt="X5"):
...
+
+        field = np.zeros((_ref.npoints, ndim))
+
+        for d in range(ndim):
+            field[:, d] = self._coeffs[..., d].reshape(-1) @ self._weights
```
effigies, line 161: Mostly a note to myself... it's hard to keep track of these

```suggestion
            #  1 x Nvox :                          (1 x K) @ (K x Nvox)
            field[:, d] = self._coeffs[..., d].reshape(-1) @ self._weights
```

nitransforms/nonlinear.py
```diff
@@ -90,70 +118,162 @@ def from_filename(cls, filename, fmt="X5"):
...
+            output_dtype=output_dtype,
+        )
+
+    def map(self, x, inverse=False):
+        r"""
+        Apply the transformation to a list of physical coordinate points.
+
+        .. math::
+            \mathbf{y} = \mathbf{x} + \Psi^3(\mathbf{k}, \mathbf{x}),
+            \label{eq:2}\tag{2}
```
effigies, line 219:
```suggestion
            \label{eq:1}\tag{1}
```

nitransforms/nonlinear.py
```diff
@@ -90,70 +118,162 @@ def from_filename(cls, filename, fmt="X5"):
...
+        # If locations to be interpolated are on a grid, use faster tensor-bspline calculation
+        if self._weights is None:
+            self._weights = grid_bspline_weights(_ref, self._knots)
+
+        field = np.zeros((_ref.npoints, ndim))
+
+        for d in range(ndim):
+            field[:, d] = self._coeffs[..., d].reshape(-1) @ self._weights
+
+        return field.astype(dtype)
```
effigies, line 163: Seems more sensible to wrap this field with metadata for interpretation now, than to make the caller do it.

```suggestion
        return DisplacementsFieldTransform(
            field.astype(dtype).reshape(*_ref.shape, -1), reference=_ref)
```

nitransforms/nonlinear.py
```diff
@@ -90,70 +118,162 @@ def from_filename(cls, filename, fmt="X5"):
...
+                cval=cval,
+                prefilter=prefilter,
+                output_dtype=output_dtype,
+            )
+
+        # If locations to be interpolated are on a grid, generate a displacements field
+        return DisplacementsFieldTransform(
+            self.to_field().reshape((*(_ref.shape), -1)),
+            reference=_ref,
+        ).apply(
```
effigies, line 203:
```suggestion
        return self.to_field().apply(
```

effigies (approved):
Based on your comment, I didn't really try to understand `_map_xyz`. The remaining parts seem sensible. Some final comments.

nitransforms/nonlinear.py
oesteban: Yes, perhaps this should be more explicit that field must be an image if reference is `None`. Will add an error

nitransforms/nonlinear.py
oesteban: The problem is that `_ref` might be `None` and it should be possible to still get the displacements.

nitransforms/nonlinear.py
oesteban: Is this a comment on the function's name, the function's behavior, or both? 
Happy to move towards something more appropriate.

nitransforms/interp/bspline.py
oesteban: *i* indexes the dimension (therefore 1, 2, 3 subscripts). *N* is the number of voxels (reference.dataobj.size) and *K* the number of control points (knots.dataobj.size). I have deeply re-written the equation to make it clearer.

Issue #40, opened 2019-10-30
Revise XFailed tests


oesteban:
Right now we have a good bunch of tests allowed to fail, and I think some of them can be already fixed to pass.

oesteban:
#145 removes xfail marks - closing to clear up my mind.

PR #142, opened 2022-02-15
FIX: Load arrays of linear transforms from FSL files

oesteban:
Implements importing series of textfiles (<name>.000, ...) generated by FSL tools (e.g., MCFLIRT).

References: #40.

### DIFF
--- a/nitransforms/io/fsl.py
+++ b/nitransforms/io/fsl.py
@@ -76,14 +76,14 @@ class FSLLinearTransform(LinearParameters):
 
     def to_ras(self, moving=None, reference=None):
         """Return a nitransforms internal RAS+ matrix."""
+        if reference is None:
+            raise ValueError("Cannot build FSL linear transform without a reference")
+
         if moving is None:
             warnings.warn(
                 "Converting FSL to RAS: moving image not provided, using reference."
             )
             moving = reference
-
-        if reference is None:
-            raise ValueError("Cannot build FSL linear transform without a reference")
 
         reference = _ensure_image(reference)
         moving = _ensure_image(moving)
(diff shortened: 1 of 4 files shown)


nitransforms/io/fsl.py
```diff
@@ -55,6 +69,35 @@ def from_string(cls, string):
...
+        # Adjust for moving image offset and orientation
+        movswp, movspc = _fsl_aff_adapt(moving)
+        post = np.linalg.inv(movswp).dot(movspc.dot(np.linalg.inv(moving.affine)))
+
+        mat = self.structarr["parameters"].T
+
+        return (
+            np.linalg.inv(post)
+            @ np.swapaxes(np.linalg.inv(mat), 0, 1)
+            @ np.linalg.inv(pre)
```
effigies, line 98: Seems to be a log of inversion.

```suggestion
        pre = refswp @ refspc @ np.linalg.inv(reference.affine)

        # Adjust for moving image offset and orientation
        movswp, movspc = _fsl_aff_adapt(moving)
        post = moving.affine @ movspc @ np.linalg.inv(movswp)

        mat = self.structarr["parameters"].T

        return post @ np.swapaxes(np.linalg.inv(mat), 0, 1) @ pre
```

I'll note that pre- and post- are not fully symmetric, as I would have expected `movspc` to be inverted the same as `movswp`. If that's a bug:

```suggestion
        pre = refswp @ refspc @ np.linalg.inv(reference.affine)

        # Adjust for moving image offset and orientation
        movswp, movspc = _fsl_aff_adapt(moving)
        post = moving.affine @ np.linalg.inv(movspc) @ np.linalg.inv(movswp)

        mat = self.structarr["parameters"].T

        return post @ np.swapaxes(np.linalg.inv(mat), 0, 1) @ pre
```

Just another note: with this much use, I'd probably just `from numpy.linalg import inv` instead of specifying it fully every time.

oesteban:
Done both suggestions - thanks!

PR #143, opened 2022-02-15
FIX: Load arrays of linear transforms from AFNI files

oesteban:
Implements importing text files with series of transforms generated by `3dVolreg`.

References: #40.


nitransforms/io/afni.py
```diff
@@ -84,6 +92,26 @@ def from_string(cls, string):
         sa["parameters"] = parameters
         return tf
 
+    def to_ras(self, moving=None, reference=None):
+        """Return a nitransforms internal RAS+ matrix."""
+        pre = LPS.copy()
+        post = LPS.copy()
```
effigies, line 98: Is this preparing to do something to them?

PR #144, opened 2022-02-15
FIX: Increase FSL serialization precision

effigies:
Was reviewing when you merged. :-)


nitransforms/io/fsl.py
```diff
@@ -46,14 +46,14 @@ def from_ras(cls, ras, moving=None, reference=None):
 
         # Adjust for reference image offset and orientation
         refswp, refspc = _fsl_aff_adapt(reference)
-        pre = reference.affine.dot(inv(refspc).dot(inv(refswp)))
+        pre = reference.affine @ inv(refswp @ refspc)
```
effigies, line 49: I think it makes sense to think of the matrices as `aff` and `swp @ spc`. So you are always either `aff @ inv(swp @ spc)` or `swp @ spc @ inv(aff)`.

nitransforms/io/fsl.py
```diff
@@ -46,14 +46,14 @@ def from_ras(cls, ras, moving=None, reference=None):
...
 
         # Adjust for moving image offset and orientation
         movswp, movspc = _fsl_aff_adapt(moving)
-        post = inv(movswp).dot(movspc.dot(inv(moving.affine)))
+        post = movswp @ movspc @ inv(moving.affine)
```
effigies, line 53: That catches that here you had `movswp` inverted and `movspc` not.

oesteban (approved):
Thanks for this. I've updated the test files so that the new precision is honored.

nitransforms/tests/utils.py
```diff
@@ -24,4 +24,4 @@ def assert_affines_by_filename(affine1, affine2):
     else:
         xfm1 = np.loadtxt(str(affine1))
         xfm2 = np.loadtxt(str(affine2))
-        np.testing.assert_almost_equal(xfm1, xfm2)
+        np.allclose(xfm1, xfm2, atol=1e-04)
```
effigies, line 27: We just lost an assertion.

PR #145, opened 2022-02-16
FIX: Refactor of LTA implementation

oesteban:
This PR makes the LTA i/o implementation more consistent with the others, cleaning up some bugs, completing some missing aspects of the implementation (which help reduce the XFailed tests by 4).

Related: #40.


nitransforms/io/lta.py
```diff
@@ -251,15 +278,40 @@ def from_string(klass, string):
...
+        # However, the affine needs to be inverted
+        # (i.e., it is not a pure "points" convention).
+        # This inversion is consistent with self.to_ras()
+        sa["m_L"] = np.linalg.inv(ras)
+        # to make LTA file format
+        return lt
+
 
-class LinearTransformArray(BaseLinearTransformList):
+class FSLinearTransformArray(BaseLinearTransformList):
```
effigies, line 307: Are we conflating FreeSurfer LTAs with something like a per-volume list of affines that one might use to motion-correct a BOLD series?

oesteban: Opened #147. @effigies - would you be okay addressing this on a separate PR?

PR #164, opened 2022-05-02
MAINT: Loosen dependencies

mgxd:


oesteban:
Hi @mgxd, sorry for my slow coming back to this. Could you remind me why this is necessary? It doesn't seem like a good idea to loosen up restrictions if the versions installed of numpy or scipy are not supported by the Python version.

effigies:
Just looking at this... You really don't want a base library to have complicated install-time dependencies. Just declare a minimum version and move on. I suspect this was used when we were using `python setup.py` instead of `pip`, and the dependency resolution was terrible and did not respect `python_requires`.

oesteban:
We probably want to update pins though, before going ahead, right?

mgxd:
I think it should be fine, I'll rebase and trigger tests again

.readthedocs.yaml
```diff
@@ -0,0 +1,19 @@
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+version: 2
+
+build:
+  os: ubuntu-20.04
+  tools:
+    python: "3.8"
```
effigies, line 9:
```suggestion
  os: ubuntu-22.04
  tools:
    python: "3.11"
```

setup.cfg
```diff
@@ -22,10 +22,8 @@ url = https://github.com/nipy/nitransforms
 [options]
 python_requires = >= 3.7
```
effigies, line 23:
```suggestion
python_requires = >= 3.8
```

effigies (approved):
Fine with me. We could also tweak these, or I can open a new PR after this is in.

effigies:
@oesteban Can you make me an admin on ReadTheDocs? I want to set up PR builds.

oesteban:
> @oesteban Can you make me an admin on ReadTheDocs? I want to set up PR builds.

done!

effigies:
Sorry, Mathias. I forgot about dropping 3.7 in CI. Looks like we haven't been testing 3.10 or 3.11 either. Happy to take this over if you don't want to babysit the CI loops.

mgxd:
Sure, I think this can be merged so you can just build off master

effigies:
Sounds good.

PR #165, opened 2022-07-18
ENH: API change in ``TransformChain`` - new composition convention

### LINKED ISSUE #81: Ordering of tranforms in TransformChain
In the context of #79, one issue has surfaced. We need to set some convention for the ordering of transforms when concatenating transforms.

At the moment, for individual transforms, we are trying to hold on to the "points" criteria. If we want to keep that for transform chains, then the order of the composed transforms should be reversed to the "image" criteria. In other words, say we have a mapping calculated with ANTs where we have 3 affine registration levels and one final nonlinear layer. Nipype and antsApplyTransforms will expect the ordering to be images (as in, we first apply the 3 affines to the data and finally the nonlinear). But coordinates are mapped the other way around (i.e., the first transform applied would be the nonlinear, and then the three linear).

I would prefer the "points" criteria, although that could fill counter-intuitive w.r.t. the existing software.

EDIT: Actually, I've just found that this problem is present in our preprint (https://osf.io/8aq7b). The last line of column 1 in page 2 says
.

Here, the order of transforms is wrong, it should actually say
.

oesteban:
This PR makes the ``TransformChain`` class more consistent with the
overall coordinate system, assuming that transforms are chained with the
*points* criteria.

In other words, in the typical setup where we have estimated one
initializing affine and then perhaps two levels of nonlinear
deformations, when calculating the coordinates of a given index in the
reference image, the last nonlinear should be applied first, then the
second, and finally the affine to pull information from the moving
image.

In other words, the chaining (composition) operation works exactly as
a single transformation procedure.

Resolves #81.

### DIFF
--- a/nitransforms/linear.py
+++ b/nitransforms/linear.py
@@ -13,6 +13,7 @@ from pathlib import Path
 from scipy import ndimage as ndi
 
 from nibabel.loadsave import load as _nbload
+from nibabel.affines import from_matvec
 
 from nitransforms.base import (
     ImageGrid,
@@ -217,6 +218,24 @@ should be (0, 0, 0, 1), got %s."""
         raise TransformFileError(
             f"Could not open <{filename}> (formats tried: {', '.join(fmtlist)})."
         )
+
+    @classmethod
+    def from_matvec(cls, mat=None, vec=None, reference=None):
+        """
+        Create an affine from a matrix and translation pair.
+
+        Example
+        -------
+        >>> Affine.from_matvec(vec=(4, 0, 0))  # doctest: +NORMALIZE_WHITESPACE
+        array([[1., 0., 0., 4.],
+               [0., 1., 0., 0.],
+               [0., 0., 1., 0.],
+               [0., 0., 0., 1.]])
+
+        """
+        mat = mat if mat is not None else np.eye(3)
+        vec = vec if vec is not None else np.zeros((3,))
+        return cls(from_matvec(mat, vector=vec), reference=reference)
 
     def __repr__(self):
         """
--- a/nitransforms/manip.py
+++ b/nitransforms/manip.py
@@ -131,8 +131,8 @@ class TransformChain(TransformBase):
             raise TransformError("Cannot apply an empty transforms chain.")
 
         transforms = self.transforms
-        if not inverse:
-            transforms = self.transforms[::-1]
+        if inverse:
+            transforms = list(reversed(self.transforms))
 
         for xfm in transforms:
             x = xfm(x, inverse=inverse)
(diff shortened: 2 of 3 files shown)


oesteban:
> If I'm getting this right, we had previously adopted the ITK last-in-first-out ordering of transform chains (I'm guessing this was because our exemplar of transform chains occurring within a file.) and now we would like it to be more intuitively left-to-right, moving-to-fixed?

Yes, I think this is a good summary -- although the convention proposed in the PR I believe should be called *right-to-left*, *fixed-to-moving*, as in the actual transformation equations:

$$
\mathbf{x}' = T_\text{2}(T_\text{1}(T_\text{0}(\mathbf{x})))
$$

where $\mathbf{x}$ is some $(x, y, z)$ coordinate in the fixed image's domain $\mathcal{F}$ and $\mathbf{x}'$ denotes the corresponding coordinate in the moving domain $\mathcal{M}$.



> is this more of an API change than a fix?

Yes.



> I would also use a different test for chaining. You currently are just showing that two inverse matrices cancel out, when something that will fail if we get the order wrong would be something like:

I'll add your test :)

That said, there is a test that failed when I just updated the API - see https://github.com/nipy/nitransforms/pull/165/commits/7f1657c9841127a2b3a8eae313670f231942ea16

The test failed because I hadn't updated it to the new ordering.

> Speaking of which, since you have two implementations of the ordering, I think we need to verify consistency:

If there are two different implementations of the ordering, that's my bad -- that shouldn't be. However, now I realize that the `asaffine()` feature was a bad idea and should probably disappear.

Instead, there should be just a `compose()`, `consolidate()` or `collapse()` (#89) that is smart enough to identify if there are nonlinear components and build an affine if there aren't.

effigies:
> > If I'm getting this right, we had previously adopted the ITK last-in-first-out ordering of transform chains (I'm guessing this was because our exemplar of transform chains occurring within a file.) and now we would like it to be more intuitively left-to-right, moving-to-fixed?
> 
> Yes, I think this is a good summary -- although the convention proposed in the PR I believe should be called _right-to-left_, _fixed-to-moving_, as in the actual transformation equations:
> 
> x′=T2(T1(T0(x)))
> 
> where x is some (x,y,z) coordinate in the fixed image's domain F and x′ denotes the corresponding coordinate in the moving domain M.
> 
> EDIT: forgot to mention that the subscript i matches python indexing, and the `transforms` property of `TransformChain` then holds [T0,T1,T2] in that order. Conversely, the ITK convention would be to store [T2,T1,T0]

With ITK, it does a stack, e.g., `[norm, coreg]`. In your formulation above, you would write `x' = coreg(norm(x))` to get the location in the moving image corresponding to the target location in the fixed image. This corresponds to ITK holding `[T0, T1]`, unless there's something different about how they map values inside the H5 files?

oesteban:
> With ITK, it does a stack, e.g., `[norm, coreg]`. In your formulation above, you would write `x' = coreg(norm(x))` to get the location in the moving image corresponding to the target location in the fixed image. This corresponds to ITK holding `[T0, T1]`, unless there's something different about how they map values inside the H5 files?

Just opened the issue above to make sure this does not fall through the cracks.

Issue #169, opened 2022-07-20
Ensure composition should be implemented with ``__matmul__`` and not ``__rmatmul__``


oesteban:
As brought up by Chris in the last bit of https://github.com/nipy/nitransforms/pull/166#issuecomment-1189631982, the right implementation should be carefully considered.

Issue #167, opened 2022-07-19
Double-check ITK's `.h5` composites


oesteban:
Ensure we have tests of `.h5` composites to make sure that #81 was properly addressed (see https://github.com/nipy/nitransforms/pull/165#issuecomment-1189201198 for further details).

PR #170, opened 2022-07-20
ENH: Collapse linear and nonlinear transforms chains

### LINKED ISSUE #89: Extend the method to "collapse" transforms in a ``TransformChain`` to the nonlinear case

oesteban:
Very undertested, but there is a new test that uses a "collapsed" transform from an ITK's .h5 file with one affine and one nonlinear (and it works).

BSpline transforms are not currently supported.

Related: #167, #169.
Resolves #89.


nitransforms/manip.py
```diff
@@ -8,7 +8,6 @@
 ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Common interface for transforms."""
 from collections.abc import Iterable
```
effigies, line 10:
```suggestion
from collections.abc import Iterable
from functools import reduce
import operator as op
```

nitransforms/manip.py
```diff
@@ -178,9 +177,8 @@ def asaffine(self, indices=None):
             The indices of the values to extract.
 
         """
-        affines = self.transforms if indices is None else np.take(self.transforms, indices)
-        retval = affines[0]
-        for xfm in affines[1:]:
+        retval = self.transforms[-1]
+        for xfm in reversed(self.transforms[:-1]):
             retval = xfm @ retval
         return retval
```
effigies, line 183: I feel like it would be more intuitive to swap the arguments of the `@` than to reverse the order of the list:

```suggestion
        retval = affines[0]
        for xfm in affines[1:]:
            retval = retval @ xfm
        return retval
```

But we can also just use a reduce (I've added the imports above if you want to go this way):

```suggestion
        return reduce(op.matmul, self.transforms)
```

nitransforms/tests/test_linear.py
```diff
@@ -328,10 +328,10 @@ def test_mulmat_operator(testdata_path):
     mat2 = from_matvec(np.eye(3), (4, 2, -1))
     aff = nitl.Affine(mat1, reference=ref)
 
-    composed = aff @ mat2
+    composed = aff @ nitl.Affine(mat2)
     assert composed.reference is None
-    assert composed == nitl.Affine(mat1.dot(mat2))
+    assert composed == nitl.Affine(mat2 @ mat1)
```
effigies, line 333: Can you add comment on why we should expect `Affine(mat1) @ Affine(mat2) == Affine(mat2 @ mat1)`? This seems counterintuitive.

effigies (commented):
Apologies for the slow response. I'm worried that this is doing something backwards driven by a misinterpretation of ITK's H5, rather than correcting an internal representation. I suspect what needs to happen is reversing the order of ITK's list when we take it in. From an API perspective, it seems almost guaranteed to trip up users if `Aff(m1) @ Aff(m2) != Aff(m1 @ m2)`.

oesteban:
@mattcieslak also this (I'm remembering as I go) :D

PR #174, opened 2023-03-29
ENH: Read ITK's composite transforms with only affines

oesteban:
Enables reading the "collapsed" output of ANTs (when no nonlinear components are inside)


effigies:
I assume we could include a pretty small example file to thoroughly exercise this?

oesteban:
> I assume we could include a pretty small example file to thoroughly exercise this?

Yeah, we should :P -- I'll add one ASAP.

oesteban:
Added tests - merging in :)

Issue #175, opened 2023-04-28
Resampled images do not properly set x-forms


oesteban:
They are copied from the moving image and that is not appropriate.

PR #176, opened 2023-04-28
FIX: Set x-forms on resampled images

oesteban:
Resolves: #175.


oesteban:
Tests are failing for unrelated reasons (#177).

PR #181, opened 2023-07-10
MNT: Drop Python 3.7 support, test through 3.11

effigies:
Some additional cleanups. Not touching CircleCI right now, as it doesn't seem broken and it's at least testing 3.9.

Also ran `pyupgrade --py38-plus`. Automated housekeeping seems like a win.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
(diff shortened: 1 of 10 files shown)


effigies:
This should be uncontroversial. OTOH, dropping a Python version should probably result in a minor release, so I don't know if you want to wait until you're ready to make 23.0.x releases require a maintenance branch.

mgxd:
As this package was created as a stopgap prior to nibabel integration, I'm not convinced there should be any maintenance branches needed - we can always just cut 23.1.0 if needed

mgxd (approved):

PR #182, opened 2023-08-09
FIX: Output displacement fields

bpinsard:
Write functions to output displacement fields from nitransforms to nifti for ITK/FSL/AFNI. 
see https://github.com/nipreps/niworkflows/pull/819#discussion_r1287485844

### DIFF
--- a/nitransforms/io/afni.py
+++ b/nitransforms/io/afni.py
@@ -193,6 +193,17 @@ class AFNIDisplacementsField(DisplacementsField):
 
         return imgobj.__class__(field, imgobj.affine, hdr)
 
+    @classmethod
+    def to_image(cls, imgobj):
+        """Export a displacements field from a nibabel object."""
+
+        hdr = imgobj.header.copy()
+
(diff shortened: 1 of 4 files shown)


nitransforms/io/afni.py
```diff
@@ -193,6 +193,17 @@ def from_image(cls, imgobj):
 
         return imgobj.__class__(field, imgobj.affine, hdr)
 
+    @classmethod
+    def to_image(cls, imgobj):
+        """Export a displacements field from a nibabel object."""
+
+        hdr = imgobj.header.copy()
+
+        warp_data = imgobj.get_fdata().reshape(imgobj.shape[:3] + (1, imgobj.shape[-1]))
```
oesteban, line 202: @effigies, shouldn't the dataobj be reshaped directly here without get_fdata?

effigies: If we were only reshaping, sure. But since we need to flip signs in the next line, you do need the loaded array.

oesteban (approved):

PR #186, opened 2023-11-16
FIX: Inefficient iterative reloading of reference and moving images

oesteban:
The list comprehension had ``nb.load``s and other method calls continously happening when converting the affine array into RAS.


nitransforms/io/afni.py
```diff
@@ -130,9 +137,21 @@ class AFNILinearTransformArray(BaseLinearTransformList):
...
+
+        post_rotation = None
+        if moving is not None:
+            mov_aff = _ensure_image(moving).affine
+            post_rotation = _cardinal_rotation(mov_aff, True) if _is_oblique(mov_aff) else None
+
+        return np.stack([
+            xfm.to_ras(pre_rotation=pre_rotation, post_rotation=post_rotation)
+            for xfm in self.xforms
+        ])
```
effigies, line 154: I would find it cleaner not to pass this up into `.to_ras()`, but instead to make your adjustments directly here:

```suggestion
        pre_rotation = post_rotation = np.eye(4)
        
        if reference is not None and _is_oblique(ref_aff := _ensure_image(reference).affine):
            pre_rotation = _cardinal_rotation(ref_aff, True)
        if moving is not None and _is_oblique(mov_aff := _ensure_image(moving).affine):
            post_rotation = _cardinal_rotation(mov_aff, True)

        return np.stack([
            post_rotation @ xfm.to_ras() @ pre_rotation
            for xfm in self.xforms
        ])
```

oesteban: way cleaner :)

nitransforms/io/afni.py
```diff
@@ -108,17 +108,24 @@ def from_string(cls, string):
         sa["parameters"] = parameters
         return tf
 
-    def to_ras(self, moving=None, reference=None):
+    def to_ras(self, moving=None, reference=None, pre_rotation=None, post_rotation=None):
```
effigies, line 111: I would just revert the changes to this function. You could unify it with what we added below, if you want to spend the effort.

nitransforms/io/afni.py
```diff
@@ -144,14 +152,22 @@ def to_string(self):
...
     def from_ras(cls, ras, moving=None, reference=None):
         """Create an ITK affine from a nitransform's RAS+ matrix."""
         _self = cls()
+
+        pre_rotation = post_rotation = np.eye(4)
+
+        if reference is not None and _is_oblique(ref_aff := _ensure_image(reference).affine):
+            pre_rotation = _cardinal_rotation(ref_aff, False)
+        if moving is not None and _is_oblique(mov_aff := _ensure_image(moving).affine):
+            post_rotation = _cardinal_rotation(mov_aff, True)
```
effigies, line 167: Looks like you have these `True`/`False`s reversed from above.

effigies: Ah, nevermind, just saw this is `from_ras`.

Issue #173, opened 2023-02-14
Memory issues when resampling large 4D files


oesteban:
The resampling of transform-mappings is typically killed in standard settings when files are large (we've seen it with DWIs of ~150 volumes).

oesteban:
@effigies - would you agree that:

1. We want to open only one volume of the reference image and then make sure it is cleared out from memory
2. We want to open the moving image with memory mapping
3. We want to resample volume-by-volume (or by N-length 4D chunks), so that the full file is not required in memory at once.

oesteban:
For 1, I think if we open memory-maping it should be sufficient as we don't access the data array, correct?

https://github.com/nipy/nitransforms/blob/1674e86a73595356eb6a775fd5b5c612952482a0/nitransforms/linear.py#L440

For 2, the moving image is loaded here:

https://github.com/nipy/nitransforms/blob/1674e86a73595356eb6a775fd5b5c612952482a0/nitransforms/linear.py#L447

and then this is the relevant loop where we should make sure only one volume is loaded on to memory at a time:

https://github.com/nipy/nitransforms/blob/1674e86a73595356eb6a775fd5b5c612952482a0/nitransforms/linear.py#L466-L476

does it make sense?

effigies:
1) You do not need to load a volume of the reference image at all, unless you're using a mask to narrow the field-of-view. All you need is the shape, affine and dtype.
2) If you switch to using `dataobj`, then it doesn't matter whether you have a mmap or not. You can create an iterator that returns one volume at a time:

    ```python
    (dataobj[:, :, :, i].astype(dtype, copy=False) for i in range(dataobj.shape[3]))
    ```

   This could be made into a function if you need more flexibiilty.

3) Volume-by-volume definitely makes sense. Let the thread manager choose how many to load, and then set your threads based on available cores and memory.

oesteban:
Thanks SO MUCH :)

effigies:
I think you have a serious inefficiency here: https://github.com/nipy/nitransforms/blob/1674e86a73595356eb6a775fd5b5c612952482a0/nitransforms/linear.py#L464-L478

In fMRIPrep, I've moved to

```python
    # Order F ensures individual volumes are contiguous in memory
    # Also matches NIfTI, making final save more efficient
    out_array = np.zeros(coordinates.shape[1:] + data.shape[-1:], dtype=output_dtype, order='F')

    tasks = [
        asyncio.create_task(
            worker(
                partial(
[...]

PR #187, opened 2023-11-16
FIX: Postpone coordinate mapping on linear array transforms

oesteban:
This PR transforms the coordinates of the reference into the moving image for each timepoint separately, effectively minimizing the memory footprint.

Resolves: #173.

### DIFF
--- a/nitransforms/linear.py
+++ b/nitransforms/linear.py
@@ -436,6 +436,7 @@ class LinearTransformsMapping(Affine):
             The data imaged after resampling to reference space.
 
         """
+
         if reference is not None and isinstance(reference, (str, Path)):
             reference = _nbload(str(reference))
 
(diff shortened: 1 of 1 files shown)


nitransforms/linear.py
```diff
@@ -446,40 +447,53 @@ def apply(
         if isinstance(spatialimage, (str, Path)):
             spatialimage = _nbload(str(spatialimage))
 
-        data = np.squeeze(np.asanyarray(spatialimage.dataobj))
-        output_dtype = output_dtype or data.dtype
+        # Avoid opening the data array just yet
+        input_dtype = spatialimage.header.get_data_dtype()
```
effigies, line 451: This doesn't seem like a good idea. If you have `int16` on-disk, you're going to start resampling to `int16` without scale factors.

oesteban: Thanks! How would be the right way of not changing the input dtype unless an output_dtype is passed?

effigies: [`nb.arrayproxy.get_obj_dtype(img.dataobj)`](https://nipy.org/nibabel/reference/nibabel.arrayproxy.html#get-obj-dtype) will tell you what the effective dtype of a data object is. It will tend to be `float64` for functional data, as that's what scaling does to data types, but for unscaled integer data, it will give you back `img.get_data_dtype()`.

oesteban: Something like this?
```suggestion
        input_dtype = nb.arrayproxy.get_obj_dtype(spatialimage.dataobj)
```

effigies: Yes.

nitransforms/linear.py
```diff
@@ -446,40 +448,53 @@ def apply(
...
+            for t in range(spatialimage.shape[-1]):
+                # Map the input coordinates on to timepoint t of the target (moving)
+                ycoords = Affine(self.matrix[t]).map(xcoords.T)[..., : _ref.ndim]
+
+                # Calculate corresponding voxel coordinates
+                yvoxels = ras2vox.map(ycoords)[..., : _ref.ndim]
+
+                # Interpolate
+                resampled[..., t] = ndi.map_coordinates(
+                    spatialimage.dataobj[..., t].astype(input_dtype, copy=False),
```
oesteban, line 483: @effigies I'm getting

```
AttributeError: 'ArrayProxy' object has no attribute 'astype'
```

I guess this should be `np.asanyarray(spatialimage.dataobj[..., t], dtype=input_dtype)

oesteban:
@arokem this should solve the issue we were having in eddymotion before Summer

nitransforms/linear.py
effigies: getitem should always return an array. Are you sure it's not on a different line, where you use the whole dataobj?

oesteban: That's possible, I'll check again

nitransforms/linear.py
```diff
@@ -446,40 +452,49 @@ def apply(
...
+
+            # Calculate corresponding voxel coordinates
+            yvoxels = ras2vox.map(ycoords)[..., : _ref.ndim]
+
+            # Interpolate
+            resampled[..., t] = ndi.map_coordinates(
+                (
+                    dataobj
+                    if dataobj is not None
+                    else np.asanyarray(spatialimage.dataobj[..., t], dtype=input_dtype)
```
oesteban, line 495:
```suggestion
                    else spatialimage.dataobj[..., t].astype(input_dtype, copy=False),
```

PR #188, opened 2023-12-01
FIX: Composition of deformation fields

jbanusco:


nitransforms/nonlinear.py
```diff
@@ -165,17 +165,22 @@ def map(self, x, inverse=False):
...
                 order=3,
                 mode="constant",
-                cval=0,
+                cval=np.nan,
                 prefilter=True,
             ) for i in range(self.reference.ndim)
         )).T
 
+        # Set NaN values back to the original coordinates value = no displacement
+        new_map[np.isnan(new_map)] = x[np.isnan(new_map)]
```
effigies, line 181: It would be good to see a test that demonstrates what is going wrong, here and that this change fixes it correctly. One approach would be 5 coordinates on a line through the center of the image. The middle three are inside the field, and the outer two outside the field. The result would show two identical coordinates, two possibly minimally changed coordinates and one noticeably shifted coordinate.

oesteban: We have confirmed with ANTs this is correct. I will try to come up with a unit test to check this does what it promises. Thanks much @jbanusco!

Issue #196, opened 2024-04-17
Transforms should implement an `ndim` property


oesteban:
Something like:

```
>>> from nitransforms import linear
>>> linear.Affine().ndim
3
```
(only one timepoint)

```
>>> from nitransforms import linear
>>> linear.LinearTransformsMapping().ndim
4
```

In the future, we can imagine two more use cases:

```
>>> from nitransforms import linear
>>> linear.ComplexAffine().ndim
4
```
(only one timepoint)

```
>>> from nitransforms import linear
>>> linear.ComplexLinearTransformsMapping().ndim
5
```

cc/ @jmarabotto

oesteban:
We just realized they do. So this is going to be changed a little:

- TransformBase -> should raise a `TypeError` instead of returning the dimensionality of the reference.
- Affine -> should overshadow TransformBase's property and return `self._matrix.ndim + 1`. That will also work for `LinearTransformArray` without loss of generality.
- Nonlinear transforms -> for the moment these cannot be 3D so they should redefine the property and return `self._reference.ndim`.

PR #197, opened 2024-04-18
FIX: Update implementation of ``ndim`` property of transforms

jmarabotto:
Resolves: #196 .


commandline_for_tests.txt
jmarabotto, line 1: Accidentally added - will remove

nitransforms/tests/test_base.py
```diff
@@ -103,12 +104,19 @@ def _to_hdf5(klass, x5_root):
...
     xfm.reference = fname
-    assert xfm.ndim == 3
+    with pytest.raises(TypeError):
+        _ = xfm.ndim
     moved = xfm.apply(str(fname), reference=fname, order=0)
     assert np.all(
         imgdata == np.asanyarray(moved.dataobj, dtype=moved.get_data_dtype())
     )
 
+    #Test ndim returned by affine
```
oesteban, line 114:
```suggestion
    # Test ndim returned by affine
```

nitransforms/linear.py
```diff
@@ -143,6 +143,11 @@ def matrix(self):
         """Access the internal representation of this affine."""
         return self._matrix
 
+    @property
+    def ndim(self):
+        """Access the internal representation of this affine."""
+        return self._matrix.ndim + 1
+
```
oesteban, line 150:
```suggestion

```

nitransforms/linear.py
```diff
@@ -143,6 +143,11 @@ def matrix(self):
         """Access the internal representation of this affine."""
         return self._matrix
 
+    @property
+    def ndim(self):
+        """Access the internal representation of this affine."""
+        return self._matrix.ndim + 1
+    
     def map(self, x, inverse=False):
```
oesteban, line 151:
```suggestion
    @property
    def ndim(self):
        """Access the internal representation of this affine."""
        return self._matrix.ndim + 1

    def map(self, x, inverse=False):
```

nitransforms/linear.py
```diff
@@ -143,6 +143,11 @@ def matrix(self):
         """Access the internal representation of this affine."""
         return self._matrix
 
+    @property
+    def ndim(self):
+        """Access the internal representation of this affine."""
+        return self._matrix.ndim + 1
```
effigies, line 149: Surely `-1`? A 4x4 affine represents a transform in 3D space, and nothing intelligible I'm aware of in 5 dimensions.

Also this docstring is off.

effigies: Oh I see, it's a 2D matrix, so this works out to 3, but does so in all cases. What you actually want is `self._matrix.shape[0] - 1`

oesteban: This will not work for 2D transforms. But it does work for >2D:

- self._matrix is 4x4 in 3D transforms (+1)
- self._matrix is tx4x4 in 4D transforms (3D + t)
- self._matrix is tx4x4x2 in 5D transforms (3D + t + complex numbers).

5D could become "a thing" when we want to realign 3D+t images reconstructed with magnitude and phase (or complex). This looks more plausible than generalization to 2D images (2D, 2D+t, 2D+t+part).

oesteban: > Okay, so we're not considering e.g., a 5x5 affine matrix for transforms of 4D spaces, but a collection of 4x4 affines for a collection of 3D volumes.

Correct

> I don't really understand why you would treat the real and imaginary parts of complex numbers as being an extra dimension in this context.

The fact that this ndim could be 5D does not mean that the transform should use all the dimensions. I agree you (in principle) don't want to interpolate in between imaginary and real part.

> If you want to be really general, you could treat it as `self._matrix.shape[0] + self._matrix.ndim - 3`, which would handle collections of any number of ND affine spaces. But I'm okay keeping this simplification and assuming 3D volumes for now.

I think your implementation is better; why keep the simplification once you've written a better alternative down here?

PR #203, opened 2024-05-16
SurfaceTransform class

feilong:
Transforms data between different surface spaces.

Dylan is working on algorithms to derive the transforms.


nitransforms/surface.py
```diff
@@ -0,0 +1,124 @@
...
+            self.mat = mat
+        else:
+            self.mat = sparse.csr_array(mat)
+
+    def apply(self, x, inverse=False, normalize="element"):
+        """Apply the transform to surface data.
+
+        Parameters
+        ----------
+        x : array-like, shape (..., nv1)
```
oesteban, line 42: what is nv1? number of vertices?

nitransforms/surface.py
```diff
@@ -0,0 +1,124 @@
...
+            Data to transform.
```
oesteban, line 43: What is "data" in this context?

nitransforms/surface.py
feilong: Yes, the number of vertices. I use `nv1` for the number of vertices of input, and `nv2` for the output, given they are likely to be different.

nitransforms/surface.py
feilong: I was thinking typically either functional data (2D data matrices, e.g., response time series) or anatomical data (1D maps, e.g., `thickness`).

nitransforms/surface.py
```diff
@@ -0,0 +1,124 @@
...
+        with h5py.File(filename, "w") as out_file:
+            out_file.attrs["Format"] = "X5"
+            out_file.attrs["Version"] = np.uint16(1)
+            root = out_file.create_group("/0")
+            self._to_hdf5(root)
+
+        return filename
+
+    @classmethod
+    def from_filename(cls, filename, fmt=None):
```
oesteban, line 106: from filename would not read sphere.reg and so forth?

feilong: This one is only reading the sparse matrix from a file.
With the code Dylan's working on, we can generate the sparse matrix from the spheres.

Shotgunosine:
From the documentation for `wb_command -surface-sphere-project-unproject`:

> A surface registration starts with an input sphere, and moves
>       its vertices around on the sphere until it matches the template data.
>       This means that the registration deformation is actually represented as
>       the difference between two separate files - the starting sphere, and the
>       registered sphere.  Since the starting sphere of the registration may not
>       have vertex correspondence to any other sphere (often, it is a native
>       sphere), it can be inconvenient to manipulate or compare these
>       deformations across subjects, etc.

Based on that description, I think the next class we need is one that will represent registration deformation and contains the original sphere and the deformation sphere (in other words, mea culpa, @oesteban was right). I'm not 100% certain it will be useful for anything (maybe project-unproject?), but it is the representation of the deformation.

oesteban:
As I mentioned, I totally see why you wanted to focus on the interpolation matrix. In practical terms, it's all you care for.

But to calculate it, in one way or another you need to formalize the transform ;)

Shotgunosine:
@oesteban Happy to talk about it more, but I think the SurfaceCoordinateTransform is the formalization of the transform. Isn't it? The deformation itself isn't useful and transforming data by subtracting the coordinate would less accurate than looking up the new coordinates by index.

oesteban:
> @oesteban Happy to talk about it more, but I think the SurfaceCoordinateTransform is the formalization of the transform. Isn't it? The deformation itself isn't useful and transforming data by subtracting the coordinate would less accurate than looking up the new coordinates by index.

Yup, that's correct.

Shotgunosine:
In terms of use cases for surface transformations, here's what I can think of:
1) Taking some metric defined on the cortical surface in subject T1w space and determining what the value of that metric would be for every vertex in a reference space (`wb_command -metric-resample`).
2) Taking labels defined in a reference space and determining which vertices in subject space those labels should be applied to (`wb_command -label-resample`).
3) Creating a surface in which every vertex corresponds to a vertex from a reference surface, but with coordinates defined in the subject's T1w space (`wb_command -surface-sphere-project-unproject`).

We can kinda do 1 and 2 with the resampling method we've got now (though we've only implemented barycentric for this and they recommend adap_bary_area). I still have trouble wrapping my head around how `surface-sphere-project-unproject` works, but I feel like that's the next use case to try implementing to confirm that this representation of surface transformation is sufficient.

feilong:
Hi @Shotgunosine, I think `surface-sphere-project-unproject` is closely related to the transform @oesteban wants.
It defines the warp using two spheres (e.g., `sphere` and `sphere.reg`). If a new sphere `X` is aligned to one of them (say, `sphere.reg`), we can use the barycentric weights (between `sphere.reg` and `X`) and the coordinates of `sphere` to warp `X`. That is, computing the new coordinates of `X` when it's aligned to `sphere` space. Conceptually, it's a diffeomorphic transform defined by the two spheres.
In practice, this is hardly used. The only use case I know is warping between fsaverage and fs_LR spheres. That being said, it's good to have it for completeness.

nitransforms/surface.py
```diff
@@ -0,0 +1,531 @@
...
+
+        if isinstance(x, SurfaceMesh) or isinstance(x, str) or isinstance(x, pathlib.PurePath):
+            x = SurfaceMesh(x)
+            if not x.check_sphere():
+                raise ValueError("If x is a surface, it should be a sphere.")
+            x.set_radius()
+            rs_x = x._coords[:, 0] @ mat
+            rs_y = x._coords[:, 1] @ mat
+            rs_z = x._coords[:, 2] @ mat
+            y = SurfaceMesh.from_arrays(np.vstack([rs_x, rs_y, rs_z]).T, self.reference._triangles)
```
effigies, line 292: I'm confused why this is decomposed. Is it a mathematical requirement, or a hack to get things working, or...?

Dockerfile
```diff
@@ -57,7 +57,9 @@ RUN mkdir -p /opt/afni-latest \
         -name "3dTshift" -or \
         -name "3dUnifize" -or \
         -name "3dAutomask" -or \
-        -name "3dvolreg" \) -delete
+        -name "3dvolreg" -or \
+        -name "3dNwarpApply" \
+    ) -delete
```
oesteban, line 62:
```suggestion
    \) -delete
```

nitransforms/surface.py
```diff
@@ -0,0 +1,526 @@
...
+    __slots__ = ("_reference", "_moving")
+
+    def __init__(self, reference, moving):
+        """Instantiate a transform between two surfaces with corresponding vertices.
+         Parameters
+        ----------
+        reference: surface
+            Surface with the destination coordinates for each index.
+        moving: surface
+            Surface with the starting coordinates for each index.
```
oesteban, line 88: I feel this is defined in reverse.
```suggestion
        reference: surface
            Surface with the starting coordinates for each index.
        moving: surface
            Surface with the destination coordinates for each index.
```

nitransforms/surface.py
```diff
@@ -0,0 +1,526 @@
...
+            raise ValueError("Both surfaces for an index transform must have corresponding"
+                             " vertices.")
+
+    def map(self, x, inverse=False):
+        if inverse:
+            source = self.reference
+            dest = self.moving
+        else:
+            source = self.moving
+            dest = self.reference
```
oesteban, line 102: Reference gives typically the sampling, so like above:
```suggestion
        if not inverse:
            source = self.reference
            dest = self.moving
        else:
            source = self.moving
            dest = self.reference
```

nitransforms/surface.py
```diff
@@ -0,0 +1,526 @@
...
+    def __init__(self, reference, moving, interpolation_method='barycentric', mat=None):
+        """Initialize the resampling.
+
+        Parameters
+        ----------
+        reference: spherical surface of the reference space.
+            Output will have number of indices equal to the number of indicies in this surface.
+            Both reference and moving should be in the same coordinate space.
+        moving: spherical surface that will be resampled.
+            Both reference and moving should be in the same coordinate space.
```
oesteban, line 137: this matches my understanding of reference and moving

nitransforms/surface.py
```diff
@@ -0,0 +1,613 @@
...
+        return cls(reference, moving)
+
+class SurfaceResampler(SurfaceTransformBase):
+    """Represents transformations in which the coordinate space remains the same and the indicies
+     change.
+     To achieve surface project-unproject functionality:
+        sphere_in as the reference
+        sphere_project_to as the moving
+     Then apply the transformation to sphere_unproject_from
+      """
```
oesteban, line 195:
```suggestion
    """
    Represents transformations in which the coordinate space remains the same and the indices change.
    To achieve surface project-unproject functionality:
        sphere_in as the reference
        sphere_project_to as the moving
    Then apply the transformation to sphere_unproject_from
    """
```

nitransforms/tests/test_surface.py
```diff
@@ -0,0 +1,207 @@
...
+    fslr_sphere = SurfaceMesh(nb.load(fslr_sphere_path))
+
+    # test mesh correspondence test
+    with pytest.raises(ValueError):
+        sct = SurfaceCoordinateTransform(fslr_sphere, pial)
+
+    # test loading from filenames
+    sct = SurfaceCoordinateTransform(sphere_reg, pial)
+    sctf = SurfaceCoordinateTransform.from_filename(reference_path=sphere_reg_path,
+                                                    moving_path=pial_path)
```
oesteban, line 89: I believe this is reversed:
```suggestion
    sct = SurfaceCoordinateTransform(pial, sphere_reg)
    sctf = SurfaceCoordinateTransform.from_filename(reference_path=pial_path,
                                                    moving_path=sphere_reg_path)
```

nitransforms/tests/test_surface.py
```diff
@@ -0,0 +1,207 @@
...
+    assert np.all(sct.map(sct.moving._coords[:100], inverse=True) == sct.reference._coords[:100])
+    assert np.all(sct.map(sct.reference._coords[:100]) == sct.moving._coords[:100])
+    with pytest.raises(NotImplementedError):
+        sct.map(sct.moving._coords[0])
+
+    # test inversion and addition
+    scti = ~sct
+
+    assert scti + sct == SurfaceCoordinateTransform(pial, pial)
+    assert sct + scti == SurfaceCoordinateTransform(sphere_reg, sphere_reg)
```
oesteban, line 102:
```suggestion
    assert sct + scti == SurfaceCoordinateTransform(pial, pial)
    assert scti + sct == SurfaceCoordinateTransform(sphere_reg, sphere_reg)
```

nitransforms/tests/test_surface.py
```diff
@@ -0,0 +1,207 @@
...
+
+    sct.reference = pial
+    sct.moving = sphere_reg
```
oesteban, line 105:
```suggestion
    sct.reference = sphere_reg
    sct.moving = pial
```

nitransforms/tests/test_surface.py
```diff
@@ -0,0 +1,207 @@
...
+    assert np.all(scti.reference._coords == sct.reference._coords)
+    assert np.all(scti.reference._triangles == sct.reference._triangles)
+    assert scti == sct
+
+def test_SurfaceCoordinateTransformIO(testdata_path, tmpdir):
+    sphere_reg_path = testdata_path / "sub-sid000005_ses-budapest_acq-MPRAGE_hemi-R_space-fsLR_desc-reg_sphere.surf.gii"
+    pial_path = testdata_path / "sub-sid000005_ses-budapest_acq-MPRAGE_hemi-R_pial.surf.gii"
+    fslr_sphere_path = testdata_path / "tpl-fsLR_hemi-R_den-32k_sphere.surf.gii"
```
oesteban, line 113: Not used?

nitransforms/tests/test_surface.py
```diff
@@ -0,0 +1,207 @@
...
+
+    sct = SurfaceCoordinateTransform(sphere_reg_path, pial_path)
```
oesteban, line 115:
```suggestion
    sct = SurfaceCoordinateTransform(pial_path, sphere_reg_path)
```

nitransforms/tests/test_surface.py
```diff
@@ -0,0 +1,207 @@
...
+def test_ProjectUnproject(testdata_path):
+
+    sphere_reg_path = testdata_path / "sub-sid000005_ses-budapest_acq-MPRAGE_hemi-R_space-fsLR_desc-reg_sphere.surf.gii"
+    fslr_sphere_path = testdata_path / "tpl-fsLR_hemi-R_den-32k_sphere.surf.gii"
+    subj_fsaverage_sphere_path = testdata_path / "sub-sid000005_ses-budapest_acq-MPRAGE_hemi-R_space-fsaverage_desc-reg_sphere.surf.gii"
+    fslr_fsaverage_sphere_path = testdata_path / "tpl-fsLR_space-fsaverage_hemi-R_den-32k_sphere.surf.gii"
+    pial_path = testdata_path / "sub-sid000005_ses-budapest_acq-MPRAGE_hemi-R_pial.surf.gii"
+
+    # test project-unproject funcitonality
+    projunproj = SurfaceResampler(sphere_reg_path, fslr_sphere_path)
```
oesteban, line 130: I agree with reference and moving here

nitransforms/tests/test_surface.py
```diff
@@ -0,0 +1,207 @@
...
+    with pytest.raises(ValueError):
+        SurfaceResampler(sphere_reg_path, pial_path)
+    with pytest.raises(ValueError):
+        SurfaceResampler(pial_path, sphere_reg_path)
+
+    reference = fslr_sphere
+    moving = sphere_reg
+    # compare results to what connectome workbench produces
+    resampling = SurfaceResampler(reference, moving)
+    resampled_thickness = resampling.apply(subj_thickness.agg_data(), normalize='element')
```
oesteban, line 159: This brings the cortical thickness from the subject into the fslr sphere, correct?

oesteban (commented):
Looks good overall, thanks for putting this together.

We still disagree on what reference and moving is, but I'm happy to go ahead and swap them the day I manage to convince you of the convention :D

nitransforms/surface.py
```diff
@@ -0,0 +1,651 @@
...
+    coords : ndarray of shape (3, )
+
+    Returns
+    -------
+    (w, u, v, t) : tuple of float
+        ``w``, ``u``, and ``v`` are the weights of the three vertices of the
+        triangle, respectively. ``t`` is the scale that needs to be multiplied
+        to ``coords`` to make it in the same plane as the three vertices.
+
+    From: https://github.com/neuroboros/neuroboros/blob/f2a2efb914e783add2bf06e0f3715236d3d8550e/src/neuroboros/surface/_barycentric.pyx#L9-L47
```
oesteban, line 592:
```suggestion
    From: https://github.com/neuroboros/neuroboros/blob/\
f2a2efb914e783add2bf06e0f3715236d3d8550e/src/neuroboros/surface/_barycentric.pyx#L9-L47
```

nitransforms/base.py
```diff
@@ -88,6 +89,75 @@ def shape(self):
...
+            return
+
+        if isinstance(dataset, Cifti2Image):
+            raise NotImplementedError
+
+        raise ValueError("Dataset could not be interpreted as an irregular sample.")
+
+    def check_sphere(self, tolerance=1.001):
+        """Check sphericity of surface.
+        Based on https://github.com/Washington-University/workbench/blob/7ba3345d161d567a4b628ceb02ab4471fc96cb20/src/Files/SurfaceResamplingHelper.cxx#L503
```
oesteban, line 132:
```suggestion
        Based on https://github.com/Washington-University/workbench/blob/\
7ba3345d161d567a4b628ceb02ab4471fc96cb20/src/Files/SurfaceResamplingHelper.cxx#L503
```

setup.cfg
```diff
@@ -27,6 +27,7 @@ install_requires =
     scipy >= 1.6.0
     nibabel >= 3.0
     h5py
+    pathlib
```
shnizzedy, line 30: I know this has been merged in for a while, but why include `pathlib` in `install_requires` when https://github.com/nipy/nitransforms/blob/f69faaf44a9cab39141df6f546394e862ed408d2/setup.cfg#L24 and pathlib was incorporated into the standard Python library in v3.4? This requirement overrides the current-Python-version's `pathlib` with the version on PyPI for Python 3.3, which breaks `pip` (at least in C-PAC's current release Dockerfile, which [includes](https://github.com/FCP-INDI/C-PAC/blob/96db8b0b65ab1d5f55fb3b895855af34d72c17e4/requirements.txt#L36)

```pip-requirements
sdcflows==2.4.0
```

which [requires](https://github.com/nipreps/sdcflows/blob/188bc87a79e89002fbccd13a4c701e3bd2224822/requirements.txt#L7)
```pip-requirements
nitransforms>=21.0.0
```
)

effigies:
@shnizzedy I agree, this should be removed. Could you open a PR?

Issue #205, opened 2024-05-16
CircleCI tests are broken


oesteban:
- ubuntu machine image is deprecated
- python version is too old

PR #206, opened 2024-05-16
MAINT: Update CircleCI's infrastructure (machine image and Python version in Docker image)

oesteban:
Resolves: #205


.circleci/config.yml
```diff
@@ -49,10 +46,7 @@ jobs:
       - run:
           name: Install test data from GIN
           command: |
-              export PY3=$(pyenv versions | grep '3\.' |
-                           sed -e 's/.* 3\./3./' -e 's/ .*//')
-              pyenv local $PY3
-              export PATH=/opt/circleci/git-annex.linux:$PATH
+              export PATH=/opt/circleci/.pyenv/versions/3.12.2/bin/:/opt/circleci/git-annex.linux:$PATH
```
mgxd, line 49: Rather than fixing to a version, I think it'll be easier to just use the latest available
```suggestion
              pyenv local 3
```

Then you wouldn't need to fiddle with the cache

PR #195, opened 2024-04-05
ENH: Outsource ``apply()`` from transform objects

jmarabotto:
I have joined CHUV (Lausanne, Switzerland) for a 5 month placement and am working with Oscar Esteban on this project. We worked through this problem together as an example of how I can contribute and resolve issues, as I am somewhat new to GitHub. Eventually, this Resolves #192 - but for now, this acts as an example and further changes are needed :)

PR #208, opened 2024-05-18
FIX: Forgotten test using ``xfm.apply()``

oesteban:
I refuse to leave Boston before all CircleCI tests are green.

Completes: #195.


effigies:
Got CircleCI back to pushing statuses to nitransforms.

PR #211, opened 2024-06-22
FIX: Load ITK fields from H5 correctly

effigies:
Ports nipreps/fmriprep#3300 to nitransforms.

I've gotten ANTs to generate a warp field with a non-cardinal rotation matrix and anistropic spacing:

image

Need to verify that we're raveling the fixed parameters in the correct order. Numpy defaults to C.

### DIFF
--- a/nitransforms/io/itk.py
+++ b/nitransforms/io/itk.py
@@ -403,14 +403,19 @@ class ITKCompositeH5:
             if xfm["TransformType"][0].startswith(b"DisplacementFieldTransform"):
                 if only_linear:
                     continue
-                _fixed = np.asanyarray(xfm[f"{typo_fallback}FixedParameters"])
-                shape = _fixed[:3].astype("uint16").tolist()
-                offset = _fixed[3:6].astype("float")
-                zooms = _fixed[6:9].astype("float")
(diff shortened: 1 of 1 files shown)


effigies:
Oh no.

image

effigies:
```
In [37]: np.allclose(np.diag([-1, -1, 1, 1]) @ targetrot, nb.affines.from_matvec(np.reshape(params, (3, 3))))
Out[37]: True
```

nitransforms/io/itk.py
```diff
@@ -403,22 +403,27 @@ def from_h5obj(cls, fileobj, check=True, only_linear=False):
...
+                    -1,
                 )
                 field[..., (0, 1)] *= -1.0
                 hdr = Nifti1Header()
                 hdr.set_intent("vector")
                 hdr.set_data_dtype("float")
 
                 xfm_list.append(
-                    Nifti1Image(field.astype("float"), LPS @ affine @ LPS, hdr)
+                    Nifti1Image(field.astype("float"), affine @ LPS, hdr)
```
effigies, line 426:
```suggestion
                    Nifti1Image(field.astype("float"), LPS @ affine, hdr)
```

PR #213, opened 2024-07-19
🔧 Loaders and util for H5 and NIfTI transforms (propagation of contributions by @sgiavasis)

jmarabotto:
(Propagation of @sgiavasis 's PR to @jmarabotto 's fork. Original PR message copied below)

Started at the NiPreps hackathon at MIT in May 2024.
Thanks again to @jmarabotto and @oesteban + @effigies for catching me up to speed and helping orient me on the nitransforms latest.

✅ This PR:

Adds loaders for .h5 and NIfTI (.nii or .nii.gz) transform files.
Adds a function for .x5 reading which only calls the H5 loader for now; set to be expanded as needed.
Adds an xfm_util level to the CLI which will read in a transform file and internally represent it as an instance of the TransformBase class, forming the X5 structure.
For now, this simply prints out some information, which can be used to verify the transform was read in properly.
Moves all I/O and I/O-related package/API calls to nitransforms/io/base.py
Starts the process of keeping the internal logic of the X5 transform structures and functions clean so that it can be unit tested without any actual transform files.
⚠️ Needs attention:

Functions exist to write out the X5 structure/transform information to either HDF5 and X5 files, but these need to be updated and completed.
More cleanup of dependencies on the lower-level details from the higher-level logic.
❓ Questions:

I don't think the internal structure of the X5 hierarchy is complete or accurate. I started with @jmarabotto 's great flowchart as a guide, but I know this might have been in flux.
In keeping with the flat namespace of the HDF5/X5 structure, the key levels remain flat (i.e. TransformGroup/0/Transform but I'm not sure if some of the final levels, like /Transform, should be a new sub-level/sub-dictionary or if they should directly contain the NumPy array defining the actual transform, etc.
▶️ Next steps (that I have in mind but I defer to the nitransforms contributors):

Completion of the write-out functions.
Some validation showing that converting any transform to an .x5 transform and applying it produces the exact same result as using the original (ex. ANTs, FSL etc.) transform with its appropriate application tool.

EDIT (OE, 08/02/2024) Just to keep these around:

- [BEP014](https://docs.google.com/document/d/11gCzXOPUbYyuQx8fErtMO9tnOKC3kTWiL9axWkkILNE/edit)
- [X5 Draft](https://docs.google.com/document/d/1yk5O0QTAOXLdP9iSG3W8ta7IcMFypu2106c3Pnjfi-4/edit)


oesteban:
hey @jmarabotto, I believe you have mistakenly deleted this branch!

nitransforms/base.py
```diff
@@ -251,14 +252,57 @@ class TransformBase:
     __slots__ = (
         "_reference",
         "_ndim",
+        "_affine",
+        "_shape",
+        "_header",
+        "_grid",
+        "_mapping",
```
oesteban, line 259: This is looking like TransformBase is becoming an affine transform. I will try to get it back to a generic, abstract base class.

nitransforms/base.py
```diff
@@ -251,14 +252,57 @@ class TransformBase:
...
+        "TransformGroup/0": {
+            "Type": None,
+            "Transform": None,
+            "Metadata": None,
+            "Inverse": None,
+        },
+        "TransformGroup/0/Domain": {"Grid": None, "Size": None, "Mapping": None},
+        "TransformGroup/1": {},
+        "TransformChain": {},
+    }
```
oesteban, line 274: We probably want to make this struct some sort of config option. Is there such a concept in nibabel, @effigies?

nitransforms/base.py
```diff
@@ -316,33 +366,68 @@ def map(self, x, inverse=False):
...
+                "shape": x5_root.shape,
+                "data": x5_root[()],  # Read the data
+            }
+        elif isinstance(x5_root, h5py.Group):
+            storage[name] = {
+                "type": "group",
+                "attrs": dict(x5_root.attrs),
+                "members": {},
+            }
+
```
oesteban, line 430: This should go into specific loaders :)

nitransforms/cli.py
```diff
@@ -47,8 +49,43 @@ def cli_apply(pargs):
         cval=pargs.cval,
         prefilter=pargs.prefilter,
     )
-    moved.to_filename(pargs.out or f"nt_{os.path.basename(pargs.moving)}")
+    # moved.to_filename(pargs.out or f"nt_{os.path.basename(pargs.moving)}")
```
oesteban, line 52:
```suggestion
    moved.to_filename(pargs.out or f"nt_{os.path.basename(pargs.moving)}")
```

nitransforms/cli.py
oesteban, line 1: skipping this file for now

nitransforms/base.py
effigies: I don't think so, but what kind of config option are you thinking? I'm very wary of putting mutable global state in a library, as that just asks for nasty interactions with multiple libraries (and threads). We could think about making a context manager to update state, along with [ContextVar](https://docs.python.org/3/library/contextvars.html)s to manage state, if needed.

But possibly you're talking about something else? This looks like a place for a schema, but I haven't studied this code yet.

oesteban:
Thanks @jmarabotto and @sgiavasis for this keystone work that has allowed me to have a first functional version with generalization potential (in that, we can quickly extend the code to work with nonlinear transforms) through #241.

I've added you as coauthors in many commits of the favored PR.

I think that PR sets a comfortable base for future contributions (looking toward @sgiavasis :D, though @jmarabotto you are welcome to return over here whenever you want :) )

PR #215, opened 2024-07-23
ENH: Serialize+parallelize 4D ``apply()`` into 3D+t and add 'low memory' loading

oesteban:
Re-implements serialization (i.e., splitting a 4D transformation into 3D transformations). For time-dependent transforms (e.g., when also slice-time correcting), parallelization will require a different partition (spatial and/or temporal windowing).


oesteban:
Dropping parallelization for a future PR. This is ready for review.

nitransforms/resampling.py
```diff
@@ -7,57 +7,126 @@
...
-):
+    transform: TransformBase,
+    spatialimage: str | Path | SpatialImage,
+    reference: str | Path | SpatialImage = None,
+    order: int = 3,
+    mode: str = "constant",
+    cval: float = 0.0,
+    prefilter: bool = True,
+    output_dtype: np.dtype = None,
+    serialize_nvols: int = SERIALIZE_VOLUME_WINDOW_WIDTH,
```
effigies, line 97: There's no parameter in the docstring, and the whole concept of serialization here is non-obvious. To the best of my understanding, the idea is that you want to parallelize over volumes but only if you hit a threshold that justifies the overhead of multiprocessing.

Have you compared the performance of multiprocessing versus async workers? If you're able to async or thread, you will have much less overhead and might not need this additional concept.

effigies (commented):
Overall looks reasonable. The tests, I'm assuming, are mostly copied from their original locations? It's a lot to look over in detail. If there are new bits among large blocks of old, could you highlight them for review?

nitransforms/resampling.py
```diff
@@ -88,52 +161,156 @@ def apply(
...
-        targets,
-        output=output_dtype,
-        order=order,
-        mode=mode,
-        cval=cval,
-        prefilter=prefilter,
-    )
+        njobs = cpu_count() if njobs is None or njobs < 1 else njobs
+
+        with ProcessPoolExecutor(max_workers=min(njobs, n_resamplings)) as executor:
```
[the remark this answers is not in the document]
oesteban: Will look into this, absolutely +1 in avoiding process pool.

nitransforms/resampling.py
oesteban: Overall it is a great interpretation.

There's something not so explicit: if you want real 4D interpolation (e.g., head motion + slice timing corrections) then you want to set `serialize_nvols = np.inf` because you can't split by volume (we've talked about doing this in a moving window, and then this parameter would map onto the width of that window).

Happy to document better.

oesteban:
> are mostly copied from their original locations?

Yes, and added parameters to exercise new branches.

> It's a lot to look over in detail. If there are new bits among large blocks of old, could you highlight them for review?

Indeed, thanks for taking the time. I'll first see how to remove the process pool and then make a final pass to identify areas that may require focused attention.

Thanks a lot :)

nitransforms/resampling.py
oesteban: I have added this parameter's description in the docstring, as it was missing. LMKWYT

nitransforms/resampling.py
```diff
@@ -88,52 +227,140 @@ def apply(
...
         spatialimage = _nbload(str(spatialimage))
 
-    data = np.asanyarray(spatialimage.dataobj)
+    # Avoid opening the data array just yet
+    input_dtype = cap_dtype(get_obj_dtype(spatialimage.dataobj), dtype_width)
+
+    # Number of data volumes
+    data_nvols = 1 if spatialimage.ndim < 4 else spatialimage.shape[-1]
+    # Number of transforms: transforms chains (e.g., affine + field, are a single transform)
+    xfm_nvols = 1 if transform.ndim < 4 else len(transform)
```
oesteban, line 236: This is new in this PR. Julien introduced the concept of `transform.ndim` to differentiate 3D transforms that require four parameters (e.g., displacements fields are typically stored as 4D NIfTIs, but are still 3D transforms) from 4D transformations (e.g., head-motion correction affines).

Following the logic of the previous version, the output is a 4D array if:
- The input image is 4D and the transform is 3D (e.g., BOLD series and coregistration to T1w image), the same transform is applied to the series.
- The input image is 3D and the transform is 4D (e.g., mapping a fieldmap onto all volumes of BOLD series), the same moving image is transformed through the transforms series.
- The input image is 4D and the transform is 4D (e.g., head motion correction).

All these 4D transforms can be called "3D+t" if the coordinates in time are not moving (e.g., no slice-timing correction). If the transform is 3D+t, then we can serialize (i.e., apply one-by-one) and use an embarrassingly parallel approach for concurrence.

Julien also implemented 4D transforms -- this was new in his PR and would require more thorough testing (esp. testing moving coordinates in time). This code is executed if `serialize_4d is not True`.

oesteban:
Seems to be working and the issues have been addressed - merging.

PR #216, opened 2024-07-23
FIX: Wrong warning argument name ``level`` in ``warnings.warn``

oesteban:
Related-to: #195.
Amends: #209.


effigies (approved):
Thanks for catching this.

PR #227, opened 2024-10-10
RF: Add `resamping.apply` to top module imports

mgxd:
This PR is for convenience after #195

### DIFF
--- a/nitransforms/__init__.py
+++ b/nitransforms/__init__.py
@@ -20,6 +20,7 @@ from . import linear, manip, nonlinear, surface
 from .linear import Affine, LinearTransformsMapping
 from .nonlinear import DenseFieldTransform
 from .manip import TransformChain
+from .resampling import apply
 
 try:
     from ._version import __version__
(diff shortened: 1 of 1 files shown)


effigies:
I've been thinking that maybe `apply()` isn't a great name. `nitransforms.apply()` is at least clear, but maybe we really want `nitransforms.resample()` to be explicit about how we're applying a transform.

mgxd:
Yeah technically true - I wouldn't be opposed. Just quickly started aggregating a list of what other tools are calling this, seems it goes either way.

<details>
<summary>Other tools</summary>

```
ANTs: ApplyTransforms
FSL: ApplyXFM
Workbench: -{volume, surface, ...}-resample
FreeSurfer: vol2vol
```

</details>

I do think it should be consistently named throughout (no aliasing at the top level), so if we decided to go forward with that let's do it in a separate API breaking PR

effigies:
Also: [skimage.warp](https://scikit-image.org/docs/stable/api/skimage.transform.html#skimage.transform.warp)

PR #233, opened 2024-12-20
ENH: Allow ingestion of alternatively grouped CompositeH5s

mgxd:
`ITKCompositeH5` is quite rigid ATM, bypassing the first H5 group entirely. This can be fine in some cases, but I have a transform containing a single displacement field (generated from a composite transform and reference image) which is not being read.

This PR successfully loads the DenseFieldTransform. If this looks good, can work on tests.

### DIFF
--- a/nitransforms/io/itk.py
+++ b/nitransforms/io/itk.py
@@ -383,12 +383,20 @@ class ITKCompositeH5:
         xfm_list = []
         h5group = fileobj["TransformGroup"]
         typo_fallback = "Transform"
+
+        start_idx = 1
+        for i, group in enumerate(h5group):
+            found = h5group[group].get(f'{typo_fallback}Parameters', None)
+            if found is not None:
+                start_idx = i
+                break
+
         try:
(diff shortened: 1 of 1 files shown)


effigies:
How did you generate the h5? With ants, the first group was just metadata.

Probably good to be flexible anyway, but if we're generating with Python, we might need to be careful to emulate itk.

mgxd:
It is with ants (ApplyTransforms)

https://github.com/nipreps/nibabies/blob/86264fee0b56144d0799b29a445f84e20ea733f9/nibabies/interfaces/patches.py#L72-L73

mgxd:
After thinking about this more, this was inserted in the wrong location - it likely need a separate class (`ITKDisplacementFieldH5`?) to handle this logic, as the transform produced by `antsApplyTransform` is not a composite, but rather just a sole displacement field. Since then, I have moved on with this use-case after https://github.com/nipreps/nibabies/pull/425, so this is no longer needed.

PR #237, opened 2025-05-31
Switch from zenodo.json to CITATION.cff, add contributors

effigies:
We haven't been updating contributors properly. I think it will generally be easier to have people add themselves to CITATION.cff when they do contribute.

There isn't a distinction between creators and contributors, so that's one downside here.

Please confirm your ORCIDs, names and affiliations: @bpinsard @jbanusco @mvdoc @jmarabotto @feilong @Shotgunosine @shnizzedy @coryshain


mvdoc:
LGTM. Thanks!

coryshain:
It's nice of you to include me in the citation! My info there is right. But I also modified exactly 2 lines of code so I don't think you need to credit me. I'll leave that up to you.

PR #258, opened 2025-07-21
MNT: Execute slowest tests on schedule

### LINKED ISSUE #229: Mark tests as slow to enable separate running
Running locally (Python 3.12), these tests take >10s:

```
143.71s call     nitransforms/tests/test_surface.py::test_ProjectUnproject
61.60s call     nitransforms/tests/test_surface.py::test_SurfaceResampler
12.49s call     nitransforms/tests/test_resampling.py::test_apply_transformchain
```

On GHA (Python 3.12):

```
  273.08s call     nitransforms/tests/test_surface.py::test_ProjectUnproject
  101.45s call     nitransforms/tests/test_surface.py::test_SurfaceResampler
```

It would be good to mark these as slow, exclude from the default runs, and set up a separate CI job. This is the majority of the test runtime.

jhlegarreta:
Execute slowest tests on schedule.

Fixes #229.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -62,6 +62,7 @@ jobs:
           datalad get -J 2 -d $TEST_DATA_HOME
 
   test:
+    if: github.event_name != 'schedule'
     runs-on: ubuntu-latest
     needs: [cache-test-data]
     strategy:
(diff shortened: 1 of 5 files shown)


oesteban (commented):
I dropped 3.9 in #259 because I believe version 25.x series should drop it (and backward compatibility of #256 was really hard).

That all said, what takes the most time here is CircleCI's tests. Perhaps we can drop all the docker shebang which substantially adds to the build.

jhlegarreta:
> That all said, what takes the most time here is CircleCI's tests. Perhaps we can drop all the docker shebang which substantially adds to the build.

Please, feel free to take over; I thought that issue #229 referred to Python testing purely, so did not have a look at what are other bottlenecks. I am not sure the current approach is working well either, since I do not see the GitHub checks taking place on the commit event.

.github/workflows/tox.yml
```diff
@@ -110,6 +111,53 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
         if: ${{ always() }}
 
+
+  slow-tests:
+    if: github.event_name == 'schedule':
```
effigies, line 116:
```suggestion
    if: github.event_name == 'schedule'
```

Issue #171, opened 2022-12-14
Incorrect dimension order with HDF5 warp field transforms


feilong:
Hello,

It seems that loading composite transforms using `ITKCompositeH5` may generate incorrect warp fields data.  Here is some minimal code and data to replicate the problem:
https://www.dropbox.com/sh/9q7fkiil5acd9wy/AAD9IAK-fPIQZ6HSke67nFPea?dl=0

My understanding is that the tensor dimensions were messed up when the warp field data was loaded from the HDF5 file, somehow the first and the third dimensions were swapped.

`ants_t1_to_mniComposite.h5` was from a working directory of fMRIPrep 21.0.2, and it's identical to `sub-*_from-T1w_to-MNI152NLin2009cAsym_mode-image_xfm.h5` in the output directory.  The NIFTI file was generated using `CompositeTransformUtil --disassemble ants_t1_to_mniComposite.h5 ants_t1_to_mniComposite`.

Best,
Feilong

PR #266, opened 2025-07-24
FIX: Ordering of ITK displacements fields when reading and writing

oesteban:
Transposes data as suggested by @feilong when reading and writing ITK displacements fields.

Tests have been added to ensure round-trip consistency.

Related-to: #171.


oesteban:
@feilong apologies for the noise of the force-pushes

PR #268, opened 2025-07-25
FIX: Mapping of dense fields

### LINKED ISSUE #267: BUG: Displacements field interpolator is broken
This code: https://github.com/nipy/nitransforms/blob/5647cafe5c61bbd125aa8c7761adfe5559e9f85b/nitransforms/nonlinear.py#L195-L207 seems to be returning very wrong values.

Instead of stacking three interpolations, we should run a single 4D interpolation.

oesteban:
Resolves: #267.


oesteban:
merged into #266.

Issue #270, opened 2025-08-12
`resampling.apply` chokes when transforming niftis that have fourth dimension equal to 1


psadil:
```python
import numpy as np
import nibabel as nb
import nitransforms as nt

dataobj=np.reshape(np.arange(27, dtype=np.uint8), (3,3,3))
nii = nb.nifti1.Nifti1Image(dataobj=dataobj, affine=np.eye(4))
transform=nt.linear.Affine(nii.affine, nii)
nt.resampling.apply(transform, spatialimage=nii)
print("success!")
nii2 = nb.nifti1.Nifti1Image(dataobj=dataobj[..., np.newaxis], affine=np.eye(4))
transform=nt.linear.Affine(nii.affine, nii2)
# this will fail
nt.resampling.apply(transform, spatialimage=nii2)
```
gives
```python
success!
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[1], line 12
     10 nii2 = nb.nifti1.Nifti1Image(dataobj=dataobj[..., np.newaxis], affine=np.eye(4))
     11 transform=nt.linear.Affine(nii.affine, nii2)
---> 12 nt.resampling.apply(transform, spatialimage=nii2)

File ~/git/neuro/qcapp/django-qcapp-ratings/.venv/lib/python3.12/site-packages/nitransforms/resampling.py:333, in apply(transform, spatialimage, reference, order, mode, cval, prefilter, output_dtype, dtype_width, serialize_nvols, max_concurrent)
    330         targets = np.vstack((time_row, ijk_targets))
    331         data = np.rollaxis(data, data.ndim - 1, 0)
--> 333     resampled = ndi.map_coordinates(
    334         data,
    335         targets,
    336         order=order,
    337         mode=mode,
    338         cval=cval,
    339         prefilter=prefilter,
    340     )
    342 if isinstance(_ref, ImageGrid):  # If reference is grid, reshape
    343     hdr = (
    344         _ref.header.copy()
    345         if _ref.header is not None
    346         else spatialimage.header.__class__()
    347     )

File ~/git/neuro/qcapp/django-qcapp-ratings/.venv/lib/python3.12/site-packages/scipy/ndimage/_interpolation.py:456, in map_coordinates(input, coordinates, output, order, mode, cval, prefilter)
    454     raise RuntimeError('input and output rank must be > 0')
    455 if coordinates.shape[0] != input.ndim:
[...]

PR #279, opened 2025-08-15
FIX: Respect singleton 4th-D

oesteban:
Resolves: #270 .


effigies (approved):
Had a quick look. It looks like to me, we just don't handle the 3D transform, 4D data case.

I couldn't quickly see how to handle it more generally, and this at least covers one particular case.

PR #284, opened 2025-09-05
chore: Relax dependencies to SPEC0+1yr

effigies:
25bdb2bd had very steep minimum version upgrades, which can be awkward on downstream tools.


effigies:
Damn. Looks like we depend on something in numpy 2.0. We're running slightly faster than SPEC0...

Anyway, #271 broke the spatial grid API by transposing `ndindex` and `ndcoords`. @oesteban I suggest that after this we release a 25.1.0 and yank 25.0.1 from PyPI.

oesteban:
> Looks like we depend on something in numpy 2.0. We're running slightly faster than SPEC0

Python 3.10 is supported only by numpy 2.10 onwards

effigies:
@oesteban Not sure why CircleCI failed to build the docker image, but I slimmed that process down a lot. It shaves 20min off CircleCI runs. Possibly more once we get caching.

PR #286, opened 2025-11-01
ENH: Add transform analysis utils

### LINKED ISSUE #239: Framewise displacement calculations, CLI, and perhaps, plotting module
It'd be great to have some transform analysis module, with a CLI so, given a transform at the input, the FD is calculated (and transform parameters written out in a tsv file?)

A good starting point for this could be NiFreeze's `registration.utils` module: https://github.com/nipreps/nifreeze/blob/26dd2ad5e387f30b96ad787fdcdb7940c4412597/src/nifreeze/registration/utils.py

jhlegarreta:
Add transform analysis utils.

Transfer contents from the `NiFreeze` projects so that hey can be reused across projects requiring transform analysis:
https://github.com/nipreps/nifreeze/blob/d27ba7552bbd9095c3c13b46443d87a4b5504c4c/src/nifreeze/analysis/motion.py https://github.com/nipreps/nifreeze/blob/d27ba7552bbd9095c3c13b46443d87a4b5504c4c/src/nifreeze/data/utils.py

Add a fixture to be able to reuse a random number generator across tests.

Fixes #239.

### DIFF
--- a/nitransforms/analysis/utils.py
+++ b/nitransforms/analysis/utils.py
@@ -24,6 +24,7 @@ from typing import Tuple
 
 import nibabel as nb
 import numpy as np
+from scipy.spatial.transform import Rotation as R
 
 from nitransforms.base import TransformBase
 from nitransforms.linear import Affine
@@ -166,6 +167,33 @@ def displacements_within_mask(
     return np.linalg.norm(diffs, axis=-1)
 
 
+def euler_from_matrix(affine: np.ndarray, degrees: bool = True) -> np.ndarray:
+    """Extract XYZ Euler angles from affine or rotation matrices using SciPy.
+
+    Parameters
+    ----------
+    affine : np.ndarray
+        Array with shape (..., 4, 4) or (..., 3, 3).
+    degrees : bool, optional
+        If True, return degrees; otherwise radians.
+
+    Returns
+    -------
+    np.ndarray
+        Array of shape (..., 3), Euler angles in 'xyz' convention.
+    """
+    affine = np.asarray(affine, dtype=float)
+    if affine.shape[-2:] not in ((3, 3), (4, 4)):
+        raise ValueError("affine must end with shape (3, 3) or (4, 4).")
+
+    mats = affine[..., :3, :3]
+    batch_shape = mats.shape[:-2]
+    mats_2d = mats.reshape(-1, 3, 3)
+
+    angles = R.from_matrix(mats_2d).as_euler("xyz", degrees=degrees)
+    return angles.reshape(*batch_shape, 3)
+
+
 def extract_motion_parameters(affine: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
     """Extract translation (mm) and rotation (degrees) parameters from an affine matrix.
 
@@ -181,12 +209,8 @@ def extract_motion_parameters(affine: np.ndarray) -> Tuple[np.ndarray, np.ndarra
     """
 
     translation = affine[:3, 3]
-    rotation_rad = np.arctan2(
-        [affine[2, 1], affine[0, 2], affine[1, 0]],
-        [affine[2, 2], affine[0, 0], affine[1, 1]],
-    )
-    rotation_deg = np.rad2deg(rotation_rad)
-    return *translation, *rotation_deg
+    rotation = euler_from_matrix(affine, degrees=True)
+    return *translation, *rotation
 
 
 def sample_unit_sphere(n_points: int = 8) -> np.ndarray:
--- a/nitransforms/analysis/utils.py
+++ b/nitransforms/analysis/utils.py
@@ -62,9 +62,14 @@ def compute_fd_from_motion(
     Returns
     -------
     :obj:`~numpy.ndarray`
-        The framewise displacement (FD) asat each timepoint as the L1 norm of
+        The framewise displacement (FD) at each timepoint as the L1 norm of
         frame-to-frame displacement across translations and rotation-derived
         displacements.
+
+    Raises
+    ------
+    exc:`ValueError`
+        If ``motion_parameters`` is not a 2D array with shape ``(T, 6)``.
     """
 
     # Columns expected: [tx, ty, tz, rx, ry, rz] where rotations are in degrees
@@ -123,6 +128,10 @@ def compute_fd_from_transform(
     :obj:`float`
         The average framewise displacement (FD) for the test transformation.
 
+    Raises
+    ------
+    exc:`ValueError`
+        If ``n_vertices < 1``.
     """
     if n_vertices < 1:
         raise ValueError("n_vertices must be >= 1")
@@ -197,15 +206,20 @@ def euler_from_matrix(affine: np.ndarray, degrees: bool = True) -> np.ndarray:
 
     Parameters
     ----------
-    affine : np.ndarray
+    affine : :obj:`~numpy.ndarray`
         Array with shape (..., 4, 4) or (..., 3, 3).
-    degrees : bool, optional
+    degrees : :obj:`bool`, optional
         If True, return degrees; otherwise radians.
 
     Returns
     -------
-    np.ndarray
+   :obj:`~numpy.ndarray`
         Array of shape (..., 3), Euler angles in 'xyz' convention.
+
+    Raises
+    ------
+    :exc:`ValueError`
+        If ``affine`` does not end with shape ``(3, 3)`` or ``(4, 4)``.
     """
     affine = np.asarray(affine, dtype=float)
     if affine.shape[-2:] not in ((3, 3), (4, 4)):
@@ -254,13 +268,20 @@ def sample_unit_sphere(n_points: int = 8) -> np.ndarray:
 
     Parameters
     ----------
-    n_points : int
+    n_points : obj:`int`
         Number of points on the sphere.
 
     Returns
     -------
-    numpy.ndarray
+    :obj:`~numpy.ndarray`
         An array of shape ``(n_points, 3)`` whose rows have unit norm.
+
+    Raises
+    ------
+    :exc:`TypeError`
+        If ``n_points`` is a boolean or not an integer type.
+    :exc:`ValueError`
+        If ``n_points < 1``.
 
     Examples
     --------
--- a/nitransforms/analysis/utils.py
+++ b/nitransforms/analysis/utils.py
@@ -21,11 +21,16 @@ from scipy.stats import zscore
 from nitransforms.base import TransformBase
 
 
-RADIUS = 50.0
-"""Typical radius (in mm) of a sphere mimicking the size of a typical human brain."""
+DEFAULT_FD_RADIUS = 50.0
+"""
+Default radius (in mm) of a sphere where framewise displacements are calculated.
+The choice was proposed by
+`Power et al. (2015) <https://doi.org/10.1016/j.neuroimage.2011.10.018>`__, and it
+represents approximately the mean distance from the cerebral cortex to the center of the head.
+"""
 
 
-def compute_fd_from_motion(motion_parameters: np.ndarray, radius: float = RADIUS) -> np.ndarray:
+def compute_fd_from_motion(motion_parameters: np.ndarray, radius: float = DEFAULT_FD_RADIUS) -> np.ndarray:
     """Compute framewise displacement (FD) from motion parameters.
 
     Each row in the motion parameters represents one frame, and columns
@@ -158,7 +163,7 @@ def extract_motion_parameters(affine: np.ndarray) -> Tuple[np.ndarray, np.ndarra
     return *translation, *rotation_deg
 
 
-def identify_spikes(fd: np.ndarray, threshold: float = 2.0) -> Tuple[np.ndarray, np.ndarray]:
+def identify_spikes(fd: np.ndarray, z_threshold: float = 2.0) -> Tuple[np.ndarray, np.ndarray]:
     """Identify motion spikes in framewise displacement data.
 
     Identifies high-motion frames as timepoint exceeding a given threshold value
@@ -168,7 +173,7 @@ def identify_spikes(fd: np.ndarray, threshold: float = 2.0) -> Tuple[np.ndarray,
     ----------
     fd : :obj:`~numpy.ndarray`
         Framewise displacement data.
-    threshold : :obj:`float`, optional
+    z_threshold : :obj:`float`, optional
         Threshold value to determine motion spikes.
 
     Returns
@@ -182,7 +187,7 @@ def identify_spikes(fd: np.ndarray, threshold: float = 2.0) -> Tuple[np.ndarray,
     # Normalize (z-score)
     fd_norm = zscore(fd)
 
-    mask = fd_norm > threshold
+    mask = fd_norm > z_threshold
     indices = np.where(mask)[0]
 
     return indices, mask
--- a/nitransforms/analysis/utils.py
+++ b/nitransforms/analysis/utils.py
@@ -66,7 +66,18 @@ AFFINE_SEQ_SHAPE_ERROR = "Affine input must have shape (4, 4) or (T, 4, 4)."
 
 @dataclass(frozen=True)
 class MotionParameters:
-    """Representation for translation and rotation motion parameters."""
+    """Representation for translation and rotation motion parameters.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> params = MotionParameters(
+    ...     translations=np.zeros((3, 3)),
+    ...     rotations=np.zeros((3, 3))
+    ... )
+    >>> params.translations.shape
+    (3, 3)
+    """
 
     translations: np.ndarray
     """Translational motion parameters with shape ``(T, 3)`` in mm."""
@@ -116,6 +127,22 @@ def extract_motion_parameters(
     :exc:`ValueError`
         If ``motion_parameters`` does not have shape ``(T, 6)`` or if ``fmt``
         is unrecognized.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> raw_params = np.zeros((10, 6))
+    >>> params = extract_motion_parameters(raw_params, fmt=MOTION_FORMAT_FSL)
+    >>> params.translations.shape
+    (10, 3)
+    >>> params.rotations.shape
+    (10, 3)
+
+    >>> # Using AFNI format (rotations in degrees are converted to radians)
+    >>> afni_params = np.array([[1.0, 2.0, 3.0, 180.0, 0.0, 0.0]])
+    >>> params_afni = extract_motion_parameters(afni_params, fmt=MOTION_FORMAT_AFNI)
+    >>> np.allclose(params_afni.rotations[0, 0], np.pi)
+    True
     """
     arr = np.asarray(motion_parameters, dtype=float)
     if arr.ndim != 2 or arr.shape[1] != 6:
@@ -154,6 +181,21 @@ def affine_to_motion_params(
     -------
     :class:`~nitransforms.analysis.utils.MotionParameters`
         Structured translations and rotations.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> # Convert a single 4x4 affine matrix
+    >>> mat = np.eye(4)
+    >>> params = affine_to_motion_params(mat)
+    >>> params.translations.shape
+    (1, 3)
+
+    >>> # Convert a batch of 3 affine matrices (shape: T, 4, 4)
+    >>> batch_mat = np.tile(np.eye(4), (3, 1, 1))
+    >>> params_batch = affine_to_motion_params(batch_mat)
+    >>> params_batch.translations.shape
+    (3, 3)
     """
     # Check LinearTransformsMapping before Affine
     if isinstance(affine, LinearTransformsMapping):
@@ -201,6 +243,18 @@ def motion_params_to_affine(motion_parameters: MotionParameters) -> LinearTransf
     -------
     :class:`~nitransforms.linear.LinearTransformsMapping`
         A mapping representing the resolved sequence of affine transforms.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> # Create a MotionParameters record for 3 frames
+    >>> params = MotionParameters(
+    ...     translations=np.zeros((3, 3)),
+    ...     rotations=np.zeros((3, 3))
+    ... )
+    >>> mapping = motion_params_to_affine(params)
+    >>> len(mapping)
+    3
     """
     if not isinstance(motion_parameters, MotionParameters):
         raise TypeError(MOTION_PARAMS_INST_ERROR_MSG)
@@ -249,6 +303,17 @@ def compute_fd_from_motion(
     ------
     exc:`TypeError`
         If ``motion_parameters`` is not a :class:`MotionParameters` instance.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> params = MotionParameters(
+    ...     translations=np.array([[0.0, 0.0, 0.0], [1.0, 2.0, 3.0]]),
+    ...     rotations=np.zeros((2, 3))
+    ... )
+    >>> fd = compute_fd_from_motion(params)
+    >>> fd.shape
+    (2,)
     """
     if not isinstance(motion_parameters, MotionParameters):
         raise TypeError(MOTION_PARAMS_INST_ERROR_MSG)
@@ -307,6 +372,17 @@ def compute_fd_from_transform(
     ------
     exc:`ValueError`
         If ``n_vertices < 1``.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import nibabel as nb
+    >>> from nitransforms.linear import Affine
+    >>> img = nb.Nifti1Image(np.zeros((5, 5, 5)), np.eye(4))
+    >>> xfm = Affine()
+    >>> fd = compute_fd_from_transform(img, xfm)
+    >>> isinstance(fd, float)
+    True
     """
     if n_vertices < 1:
         raise ValueError("n_vertices must be >= 1")
@@ -360,6 +436,16 @@ def displacements_within_mask(
     :obj:`~numpy.ndarray`
         An array of displacements (in mm) for each voxel within the mask.
 
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import nibabel as nb
+    >>> from nitransforms.linear import Affine
+    >>> mask_img = nb.Nifti1Image(np.ones((3, 3, 3)), np.eye(4))
+    >>> xfm = Affine()
+    >>> disps = displacements_within_mask(mask_img, xfm)
+    >>> disps.shape
+    (27,)
     """
     # Mask data as boolean (True for voxels inside the mask)
     maskdata = np.asanyarray(mask_img.dataobj) > 0
--- a/nitransforms/analysis/utils.py
+++ b/nitransforms/analysis/utils.py
@@ -93,7 +93,8 @@ def compute_fd_from_transform(
 
     This implementation varies with respect to the original formulation by [Power2012]_
     in that the FD is computed as the average across a number of vertices sampled over the
-    sphere.
+    sphere. See :func:`~nitransforms.analysis.utils.sample_unit_sphere` for details
+    about the vertex sampling method.
 
     Parameters
     ----------
--- a/nitransforms/analysis/utils.py
+++ b/nitransforms/analysis/utils.py
@@ -62,8 +62,9 @@ def compute_fd_from_motion(
     Returns
     -------
     :obj:`~numpy.ndarray`
-        The framewise displacement (FD) as the sum of absolute differences
-        between consecutive frames.
+        The framewise displacement (FD) asat each timepoint as the L1 norm of
+        frame-to-frame displacement across translations and rotation-derived
+        displacements.
     """
 
     # Columns expected: [tx, ty, tz, rx, ry, rz] where rotations are in degrees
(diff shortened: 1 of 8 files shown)


nitransforms/utils.py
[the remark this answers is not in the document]
jhlegarreta: Removed the file.

oesteban (commented):
Thanks for the update. I just realized of a little inconsistency. Other than that, I think this is almost ready.

jhlegarreta:
This is ready to go. Merging this would allow `NiFreeze` to drop code that is general, and probably have better support as these functions may get better testing. We will need to set/bump a new min version for `NiTransforms` when a new version with these changes is released.

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,188 @@
...
+
+import nibabel as nb
+import numpy as np
+from scipy.stats import zscore
+
+from nitransforms.base import TransformBase
+
+
+RADIUS = 50.0
+"""Typical radius (in mm) of a sphere mimicking the size of a typical human brain."""
```
oesteban:
```suggestion
"""
Default radius (in mm) of a sphere where framewise displacements are calculated.
The choice was proposed by
`Power et al. (2015) <https://doi.org/10.1016/j.neuroimage.2011.10.018>`__, and it
represents approximately the mean distance from the cerebral cortex to the center of the head.
"""
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,188 @@
...
+from typing import Tuple
+
+import nibabel as nb
+import numpy as np
+from scipy.stats import zscore
+
+from nitransforms.base import TransformBase
+
+
+RADIUS = 50.0
```
oesteban:
```suggestion
DEFAULT_FD_RADIUS = 50.0
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,188 @@
...
+"""Typical radius (in mm) of a sphere mimicking the size of a typical human brain."""
+
+
+def compute_fd_from_motion(motion_parameters: np.ndarray, radius: float = RADIUS) -> np.ndarray:
```
oesteban:
```suggestion
def compute_fd_from_motion(motion_parameters: np.ndarray, radius: float = DEFAULT_FD_RADIUS) -> np.ndarray:
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,188 @@
...
+    """Compute framewise displacement (FD) from motion parameters.
```
oesteban, line 45: We should cite [Power et al. (2013)](https://doi.org/10.1016/j.neuroimage.2011.10.018) (doi:10.1016/j.neuroimage.2011.10.018) somewhere here.

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,188 @@
...
+
+    translation = affine[:3, 3]
+    rotation_rad = np.arctan2(
+        [affine[2, 1], affine[0, 2], affine[1, 0]], [affine[2, 2], affine[0, 0], affine[1, 1]]
+    )
+    rotation_deg = np.rad2deg(rotation_rad)
+    return *translation, *rotation_deg
+
+
+def identify_spikes(fd: np.ndarray, threshold: float = 2.0) -> Tuple[np.ndarray, np.ndarray]:
```
oesteban: Not sure this is something very useful that we should maintain. In any case:

```suggestion
def identify_spikes(fd: np.ndarray, z_threshold: float = 2.0) -> Tuple[np.ndarray, np.ndarray]:
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,188 @@
...
+    indices : :obj:`~numpy.ndarray`
+        Indices of identified motion spikes.
+    mask : :obj:`~numpy.ndarray`
+        Mask of identified motion spikes.
+    """
+
+    # Normalize (z-score)
+    fd_norm = zscore(fd)
+
+    mask = fd_norm > threshold
```
oesteban:
```suggestion
    mask = fd_norm > z_threshold
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,188 @@
...
+        The average framewise displacement (FD) for the test transformation.
+
+    """
+    affine = img.affine
+    # Compute the center of the image in voxel space
+    center_ijk = 0.5 * (np.array(img.shape[:3]) - 1)
+    # Convert to world coordinates
+    center_xyz = nb.affines.apply_affine(affine, center_ijk)
+    # Generate coordinates of points at radius distance from center
+    fd_coords = np.array(list(product(*((radius, -radius),) * 3))) + center_xyz
```
oesteban: BUG: the radius of these points is not 50 mm:

```
In [1]: from itertools import product
    ...: import numpy as np
    ...: 
    ...: radius = 50
    ...: center_xyz = (0, 0, 0)
    ...: 
    ...: fd_coords = np.array(list(product(*((radius, -radius),) * 3))) + center_xyz
    ...: np.linalg.norm(fd_coords, axis=1)
Out[1]: 
array([86.60254038, 86.60254038, 86.60254038, 86.60254038, 86.60254038,
       86.60254038, 86.60254038, 86.60254038])
```

cc/ @effigies in the likely case I brought over this bug from other places (nipype?, fmriprep?)

oesteban (commented):
We had a pretty hefty bug in the original implementation of nifreeze!

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,188 @@
...
+    """Identify motion spikes in framewise displacement data.
+
+    Identifies high-motion frames as timepoint exceeding a given threshold value
+    based on z-score normalized framewise displacement (FD) values.
+
+    Parameters
+    ----------
+    fd : :obj:`~numpy.ndarray`
+        Framewise displacement data.
+    threshold : :obj:`float`, optional
```
jhlegarreta:
```suggestion
    z_threshold : :obj:`float`, optional
```

nitransforms/tests/test_analysis.py
```diff
@@ -0,0 +1,156 @@
...
+        assert np.allclose(params[3:], expected_rot)
+
+
+def test_identify_spikes(request):
+    rng = request.node.rng
+
+    n_samples = 450
+
+    fd = rng.normal(0, 5, n_samples)
+    threshold = 2.0
```
jhlegarreta:
```suggestion
    z_threshold = 2.0
```

nitransforms/tests/test_analysis.py
```diff
@@ -0,0 +1,156 @@
...
+
+    expected_indices = np.asarray(
+        [5, 57, 85, 100, 127, 180, 191, 202, 335, 393, 409]
+    )
+    expected_mask = np.zeros(n_samples, dtype=bool)
+    expected_mask[expected_indices] = True
+
+    obtained_indices, obtained_mask = identify_spikes(fd, threshold=threshold)
```
jhlegarreta:
```suggestion
    obtained_indices, obtained_mask = identify_spikes(fd, z_threshold=z_threshold)
```

nitransforms/analysis/utils.py
jhlegarreta: Done in 883bf93.

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,197 @@
...
+    rotation_displacement = d_rotations * radius
+
+    # Compute FD as sum of absolute differences
+    return np.sum(np.abs(d_translations) + np.abs(rotation_displacement), axis=1)
+
+
+def compute_fd_from_transform(
+    img: nb.spatialimages.SpatialImage,
+    test_xfm: TransformBase,
+    radius: float = RADIUS,
```
jhlegarreta:
```suggestion
    radius: float = DEFAULT_FD_RADIUS,
```

nitransforms/analysis/utils.py
oesteban:
```suggestion
    e_coords = np.array(list(product(*((1, -1),) * 3)))
    e_coords = e_coords / np.linalg.norm(e_coords, axis=1)[:, None]
    fd_coords = np.vstack((np.eye(3), -1 * np.eye(3), e_xyz)) * radius + center_xyz
```

We should extract this into a function (I think DIPY has something like this?) that we can test thoroughly:

```Python
import math
import numpy as np
from itertools import product


def sample_unit_sphere(n_points=8):
    """Returns :math:`N` evenly distributed points on a unit-radius sphere.

    This function returns a **deterministic**, **quasi-uniform** point set on the
    surface of the unit sphere :math:`S^2 \\subset \\mathbb{R}^3`.

    Notes
    -----
    - There is no unique notion of "evenly distributed" for arbitrary `N` on a sphere.
      This function uses:
        * **Platonic solids** for certain small `N` (high symmetry; e.g. `N=6` gives
          the ±axis points).
        * A **Fibonacci / golden-angle spiral** otherwise (fast, simple, good coverage).

    Parameters
    ----------
    n_points : int
        Number of points on the sphere.

    Returns
    -------
    numpy.ndarray
        An array of shape ``(n_points, 3)`` whose rows have unit norm.

    Examples
    --------
    Basic shape + unit norm:

    >>> import numpy as np
    >>> X = sample_unit_sphere(10)
    >>> X.shape
    (10, 3)
    >>> np.allclose(np.linalg.norm(X, axis=1), 1.0)
    True

    Edge case:

    >>> sample_unit_sphere(0)
    Traceback (most recent call last):
    ...
    ValueError: n_points must be a positive integer

    For N=6, return the ±axis points (octahedron vertices), matching your example idea:

    >>> X = sample_unit_sphere(6)
    >>> # Each row has exactly one coordinate with magnitude 1, others 0
    >>> np.all((np.abs(X) == 1.0).sum(axis=1) == 1)
    True
    >>> np.all((np.abs(X) == 1.0).sum(axis=0) == 2)  # each axis appears twice (±)
    True

    For N=4, the tetrahedron has constant pairwise dot product -1/3 off-diagonal:

    >>> X = sample_unit_sphere(4)
    >>> D = X @ X.T
    >>> off = D[~np.eye(4, dtype=bool)]
    >>> np.allclose(off, -1/3)
    True

    For a quasi-uniform set, the second moment matrix is close to I/3, and the
    minimum angular separation is non-trivial:

    >>> X = sample_unit_sphere(200)
    >>> M = (X.T @ X) / len(X)
    >>> np.allclose(M, np.eye(3) / 3, atol=1e-3)
    True
    >>> def min_angle_rad(Y):
    ...     dots = np.clip(Y @ Y.T, -1.0, 1.0)
    ...     np.fill_diagonal(dots, 1.0)
    ...     ang = np.arccos(dots)
    ...     np.fill_diagonal(ang, np.inf)
    ...     return float(ang.min())
    >>> min_angle_rad(X) > 0.18
    True
    """
    if isinstance(n_points, (bool, np.bool_)) or not isinstance(n_points, (int, np.integer)):
        raise TypeError("n_points must be an integer")
    if n_points < 1:
        raise ValueError("n_points must be a positive integer")

    def _normalize(X):
        X = np.asarray(X, dtype=float)
        X /= np.linalg.norm(X, axis=1, keepdims=True)
        return X

    # --- Highly symmetric small-N cases (Platonic solids / degeneracies) ---
    if n_points == 1:
        return np.array([[0.0, 0.0, 1.0]])
    if n_points == 2:
        return np.array([[0.0, 0.0, 1.0],
                         [0.0, 0.0, -1.0]])
    if n_points == 4:  # tetrahedron (4 cube corners with even number of minus signs)
        X = np.array([[ 1.0,  1.0,  1.0],
                      [ 1.0, -1.0, -1.0],
                      [-1.0,  1.0, -1.0],
                      [-1.0, -1.0,  1.0]])
        return _normalize(X)
    if n_points == 6:  # octahedron: ±axes
        return np.vstack([np.eye(3), -np.eye(3)]).astype(float)
    if n_points in (8, 20):  # hexahedron (cube) & base for dodecahedron
        X = np.array(list(product([-1.0, 1.0], repeat=3)), dtype=float)
        if n_points == 8:
            return _normalize(X)

        # Dodecahedron, add 12 vertices
        extra = []
        phi = (1.0 + math.sqrt(5.0)) / 2.0
        invphi = 1.0 / phi
        for a in (-1.0, 1.0):
            for b in (-1.0, 1.0):
                extra.extend([
                    [0.0, a * invphi, b * phi],
                    [a * invphi, b * phi, 0.0],
                    [a * phi, 0.0, b * invphi],
                ])
        X = np.vstack([X, np.asarray(extra, dtype=float)])
        return _normalize(X)
    if n_points == 12:  # icosahedron
        phi = (1.0 + math.sqrt(5.0)) / 2.0
        X = []
        for a in (-1.0, 1.0):
            for b in (-1.0, 1.0):
                X.append([0.0, a, b * phi])
                X.append([a, b * phi, 0.0])
                X.append([a * phi, 0.0, b])
        return _normalize(X)

    # --- General N: Fibonacci / golden-angle spiral (deterministic quasi-uniform) ---
    i = np.arange(n_points, dtype=float)
    golden_angle = math.pi * (3.0 - math.sqrt(5.0))

    z = 1.0 - 2.0 * (i + 0.5) / n_points
    r = np.sqrt(np.maximum(0.0, 1.0 - z * z))
    theta = golden_angle * i

    X = np.column_stack((r * np.cos(theta), r * np.sin(theta), z))
    return _normalize(X)

```


EDIT: Added a suggestion of how we could implement this here. This function could also be useful for surface representations in nibabel.

jhlegarreta: There are a number of ways to create the vertices over the sphere in DIPY:
https://github.com/dipy/dipy/blob/8009c8124ee7932701a375ab105373350a5c5a38/dipy/core/sphere.py
Most notably,
https://github.com/dipy/dipy/blob/8009c8124ee7932701a375ab105373350a5c5a38/dipy/core/sphere.py#L482
https://github.com/dipy/dipy/blob/8009c8124ee7932701a375ab105373350a5c5a38/dipy/core/sphere.py#L542

oesteban:
@jhlegarreta please have one more look - I realized that the FD from transform matrices was also badly computed as it did not accept the previous transform. IMHO, we should also provide a "vectorized" version of FD such that you pass the list of transforms and gives you timeseries. Could be left for a future PR though.

@effigies added you to have an extra pair of eyes on this.

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,376 @@
...
+    # Compute FD as sum of absolute differences
+    return np.sum(np.abs(d_translations) + np.abs(rotation_displacement), axis=1)
+
+
+def compute_fd_from_transform(
+    img: nb.spatialimages.SpatialImage,
+    xfm: TransformBase,
+    xfm_prev: TransformBase | None = None,
+    radius: float = DEFAULT_FD_RADIUS,
+    n_vertices: int = 8,
```
oesteban, line 86: It would be interesting to make the output match ``compute_fd_from_motion`` for ``n_vertices = 1``.

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,376 @@
...
+) -> float:
+    """
+    Compute the framewise displacement (FD) for a given transformation.
+
+    This implementation varies with respect to the original formulation by [Power2012]_
+    in that the FD is computed as the average across a number of vertices sample of the
```
jhlegarreta:
```suggestion
    in that the FD is computed as the average across a number of vertices sampled over the
```

jhlegarreta: The method to sample the sphere should be documented or cross referenced using Sphinx markup in the docstring.

oesteban (approved):

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,376 @@
...
+
+def sample_unit_sphere(n_points: int = 8) -> np.ndarray:
+    """Returns :math:`N` evenly distributed points on a unit-radius sphere.
+
+    This function returns a **deterministic**, **quasi-uniform** point set on the
+    surface of the unit sphere :math:`S^2 \\subset \\mathbb{R}^3`.
+
+    Notes
+    -----
+    - There is no unique notion of "evenly distributed" for arbitrary `N` on a sphere.
```
jhlegarreta:
```suggestion
    - There is no unique notion of "evenly distributed" for arbitrary ``N`` on a sphere.
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,376 @@
...
+      This function uses:
+        * **Platonic solids** for certain small `N` (high symmetry; e.g. `N=6` gives
```
jhlegarreta:
```suggestion
        * **Platonic solids** for certain small ``N`` (high symmetry; e.g. ``N=6`` gives
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,376 @@
...
+    ...     X = sample_unit_sphere(n_points)
+    ...     X.shape, bool(np.allclose(np.linalg.norm(X, axis=1), 1.0))
+    ((1, 3), True)
+    ((2, 3), True)
+    ((8, 3), True)
+    ((10, 3), True)
+    ((12, 3), True)
+    ((20, 3), True)
+
+    For N=6, return the ±axis points (octahedron vertices):
```
jhlegarreta:
```suggestion
    For ``N=6``, return the ±axis points (octahedron vertices):
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,376 @@
...
+
+    >>> X = sample_unit_sphere(6)
+    >>> # Each row has exactly one coordinate with magnitude 1, others 0
+    >>> bool(np.all((np.abs(X) == 1.0).sum(axis=1) == 1))
+    True
+    >>> bool(np.all((np.abs(X) == 1.0).sum(axis=0) == 2))  # each axis appears twice (±)
+    True
+
+    For N=4, the tetrahedron has constant pairwise dot product -1/3 off-diagonal:
```
jhlegarreta:
```suggestion
    For ``N=4``, the tetrahedron has constant pairwise dot product -1/3 off-diagonal:
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,376 @@
...
+
+    >>> X = sample_unit_sphere(4)
+    >>> D = X @ X.T
+    >>> off = D[~np.eye(4, dtype=bool)]
+    >>> bool(np.allclose(off, -1/3))
+    True
+
+    For a quasi-uniform set, the second moment matrix is close to I/3, and the
```
jhlegarreta:
```suggestion
    For a quasi-uniform set, the second moment matrix is close to ``I/3``, and the
```

jhlegarreta (commented):
> IMHO, we should also provide a "vectorized" version of FD such that you pass the list of transforms and gives you timeseries. Could be left for a future PR though.

Better for a future PR. This is already better than what we have in `NiFreeze` (thanks for reviewing and the subsequent work), allows us to have a common, single authoritative location, and frees `NiFreeze` from hosting this.

Some inline comments about documentation and syntax highlighting.

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,376 @@
...
+    xfm_prev: TransformBase | None = None,
+    radius: float = DEFAULT_FD_RADIUS,
+    n_vertices: int = 8,
+) -> float:
+    """
+    Compute the framewise displacement (FD) for a given transformation.
+
+    This implementation varies with respect to the original formulation by [Power2012]_
+    in that the FD is computed as the average across a number of vertices sample of the
+    sphere.
```
jhlegarreta:
```suggestion
    sphere. See :func:`~nitransforms.analysis.utils.sample_unit_sphere` for details
    about the vertex sampling method.
```

effigies:
Out of curiosity, why are we standardizing on degrees? Apart from AFNI, everybody reports angles in radians. It seems to add unnecessary complication. Nipype also normalizes to SPM style (https://github.com/nipy/nipype/blob/8234ec318489930fa17487cf15d210420214b00a/nipype/utils/misc.py#L246-L271) so consistency within the nipy world seems desirable.

effigies:
Another conceptual question: Do we want to continue the tradition of accepting an unlabeled Nx6 matrix, or would it be a good opportunity to require the caller to consider the  it useful to have a function that accepts an Nx6 matrix, or would it make more sense to accept 2 Nx3 matrices and a radius? The reason I ask is that it would be very tempting to load an Nx6 matrix out of some transform file and pass it directly without carefully considering which are the translations and which are the rotations. I might think of the API as follows:

```py
def framewise_displacement(
    translations: np.ndarray,
    rotations: np.ndarray,
    radius: float = DEFAULT_RADIUS
) -> np.ndarray:
    """Calculate framewise displacement of motion parameters. ..."""

def extract_parameters(motion_parameters: np.ndarray, fmt: str | None=None) -> tuple[np.ndarray, np.ndarray]:
    """Extract translation and rotation parameters from a parameter array.

    This function normalizes rotation parameters to radians, if the input format is known to use degrees.

    The parameters are returned in the order expected by the `framewise_displacement` function.
    """

# I know what my matrix is:
fd = framewise_displacement(motion_parameters[:, :3], motion_parameters[:, 3:])
# I know I got it from AFNI:
afni_fd = framewise_displacement(*extract_parameters(motion_parameters, fmt='afni'))
```

We could even force people to use labels instead and say:

```py
@dataclass(kw_only=True)
class MotionParameters:
    rotations: np.ndarray
    translations: np.ndarray

def extract_parameters(motion_parameters: np.ndarray, fmt: str | None=None) -> MotionParameters: ...
def framewise_displacement(motion_parameters: MotionParameters, radius: float = DEFAULT_RADIUS) -> np.ndarray: ...
```

This would even suggest a straightforward translation between affines and motion parameters:

```py
def affineToMotionParams(aff: Affine) -> MotionParameters: ...
def motionParamstoAffine(motion_parameters: MotionParameters) -> LinearTransformsMapping: ...
```

Just a couple thoughts. Will review the PR on the assumption that you'll stick with degrees and a big matrix.

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,377 @@
...
+
+    # Compute differences between consecutive frames
+    d_translations = np.vstack([np.zeros((1, 3)), np.diff(translations, axis=0)])
+    d_rotations = np.vstack([np.zeros((1, 3)), np.diff(rotations_rad, axis=0)])
+
+    # Convert rotations from radians to displacement on a sphere
+    rotation_displacement = d_rotations * radius
+
+    # Compute FD as sum of absolute differences
+    return np.sum(np.abs(d_translations) + np.abs(rotation_displacement), axis=1)
```
effigies: Conceptually, this last line reads weirdly to me. Why add the translation displacements to the rotation displacements and then sum? It does the job, but from the perspective of understanding the method, I might write this as:

```suggestion
    translations = motion_parameters[:, :3]
    rotations = np.deg2rad(motion_parameters[:, 3:])
    
    displacements = np.hstack((
        np.diff(translations, axis=0, prepend=0),
        np.diff(rotations * radius, axis=0, prepend=0)
    ))

    # FD is the L1 norm (sum of absolute values)
    return np.linalg.norm(displacements, ord=1, axis=1)
```

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,377 @@
...
+    radius: float = DEFAULT_FD_RADIUS,
+    n_vertices: int = 8,
+) -> float:
+    """
+    Compute the framewise displacement (FD) for a given transformation.
+
+    This implementation varies with respect to the original formulation by [Power2012]_
+    in that the FD is computed as the average across a number of vertices sampled over the
+    sphere. See :func:`~nitransforms.analysis.utils.sample_unit_sphere` for details
+    about the vertex sampling method.
```
effigies, line 94: This also uses the L2 norm instead of L1. What is the reasoning for this?

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,377 @@
...
+        Extracted translation and rotation parameters.
+    """
+
+    translation = affine[:3, 3]
+    rotation_rad = np.arctan2(
+        [affine[2, 1], affine[0, 2], affine[1, 0]],
+        [affine[2, 2], affine[0, 0], affine[1, 1]],
+    )
+    rotation_deg = np.rad2deg(rotation_rad)
+    return *translation, *rotation_deg
```
effigies, line 190: This isn't obviously correct to me. There are many ways of extracting angles from a rotation matrix, but I'm surprised that you seem able to get them using only 6 entries of a 3x3 matrix.

At the least, this function needs a citation, but I would be inclined to rely on scipy.spatial.transforms or transforms3d for this functionality.

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,377 @@
...
+    ----------
+    n_points : int
+        Number of points on the sphere.
+
+    Returns
+    -------
+    numpy.ndarray
+        An array of shape ``(n_points, 3)`` whose rows have unit norm.
+
+    Examples
```
effigies, line 217: Would it be too much to ask for these docs to have plots to show the points on the unit sphere?

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,377 @@
...
+    return _normalize(X)
+
+
+def identify_spikes(
+    fd: np.ndarray, z_threshold: float = 2.0
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Identify motion spikes in framewise displacement data.
+
+    Identifies high-motion frames as timepoint exceeding a given threshold value
+    based on z-score normalized framewise displacement (FD) values.
```
effigies, line 351: Is there a citation for this one? This seems dubious to me, as z-scoring will amplify low-motion subjects' FD and suppress high-motion subjects' FD. For the purposes of censoring, absolute thresholds are often used based on voxel size.

nitransforms/analysis/utils.py
```diff
@@ -0,0 +1,377 @@
...
+
+    affine = img.affine
+    # Compute the center of the image in voxel space
+    center_ijk = 0.5 * (np.array(img.shape[:3]) - 1)
+    # Convert to world coordinates
+    center_xyz = nb.affines.apply_affine(affine, center_ijk)
+    # Generate coordinates of points at radius distance from center
+    fd_coords = sample_unit_sphere(n_points=n_vertices) * radius + center_xyz
+    # Compute the average displacement from the test transformation
+    return np.mean(np.linalg.norm(xfm.map(fd_coords) - xfm_prev.map(fd_coords), axis=-1))
```
oesteban:
```suggestion
    return np.mean(np.linalg.norm(xfm.map(fd_coords) - xfm_prev.map(fd_coords), ord=1, axis=-1))
```

nitransforms/analysis/utils.py
oesteban: Thanks for pointing this out, changing to L1

nitransforms/analysis/utils.py
oesteban: Thanks, will investigate.

nitransforms/analysis/utils.py
oesteban: Sounds good, I'll see if it doesn't involve much or defer to a future PR.

nitransforms/analysis/utils.py
oesteban: I'm inclined to drop this function altogether. The purpose is not censoring, but rather identifying the higher trees in the forest.

jhlegarreta:
Can we revisit this so that we can reduce the fronts across NiFreeze, please?

nitransforms/analysis/utils.py
jhlegarreta: See cbc29d481346610468bf5da2c3cb9fd5e3dd951b.

nitransforms/analysis/utils.py
jhlegarreta: See 18faeb80725e9ec48eba978c4f16db7503cff954.

nitransforms/analysis/utils.py
jhlegarreta: See b29f8398c98936ec9b6675e80ca91057aabc5201.

nitransforms/analysis/utils.py
jhlegarreta: Removed in b459594b89f8fc5e5233e784b17634a45eacc3c7.

jhlegarreta:
I think https://github.com/nipy/nitransforms/pull/286#issuecomment-3719915982 can be addressed in a subsequent PR.

jhlegarreta:
I'd squash commits once we are satisfied with the changes.

jhlegarreta:
Note: we are assuming that the matrices are all valid: scipy 1.17 introduced a [check parameter](https://docs.scipy.org/doc/scipy-1.17.0/reference/generated/scipy.spatial.transform.Rotation.from_matrix.html) that we are not able to use [because we only require scipy >= 1.10](https://docs.scipy.org/doc/scipy-1.16.2/reference/generated/scipy.spatial.transform.Rotation.from_matrix.html).

I think I've done all I can on this PR.

jhlegarreta:
@oesteban @effigies Can you please look at this so that we can make another step forward with the bits that we'd like to outsource to here in NiFreeze?

oesteban:
I'm looking at the RTD documentation and, in addition to see an outdated version (?) showing [``:math:`pm`axis points``](https://nitransforms--286.org.readthedocs.build/en/286/_api/analysis.html#id4), I am not seeing any pictures showing where points are located, am I looking at the wrong place?

jhlegarreta:
> I'm looking at the RTD documentation and, in addition to see an outdated version (?) showing [:math:`pm`axis points](https://nitransforms--286.org.readthedocs.build/en/286/_api/analysis.html#id4), I am not seeing any pictures showing where points are located, am I looking at the wrong place?

Math symbols have been fixed and the plots are shown now.

jhlegarreta:
Can we please review this?

jhlegarreta:
Re https://github.com/nipy/nitransforms/pull/286#issuecomment-3719915982 @effigies I implemented the API. I think I have finished my work on this here and the PR is ready to be merged. Feel free to implement improvements and to merge as appropriate as I have exhausted my bandwidth here.

PR #298, opened 2026-09-14
fix: normalize resampling coordinates to a consistent layout

### LINKED ISSUE #297: 3d resampling of 4d image crashes
Trying to implement warping of data denoised from tedana in boldref to MNI, all in-memory before a GLM. 

```
import nibabel as nib
import nitransforms

img = nib.load('./derivatives/tedana/sub-HA01/func/sub-HA01_task-painexpectancy_run-01_desc-denoised_bold.nii.gz')
mask = nib.load('derivatives/fmriprep/sub-HA01/func/sub-HA01_task-painexpectancy_run-01_space-MNI152NLin2009cAsym_desc-brain_mask.nii.gz')

boldref_to_t1 = nitransforms.linear.load('derivatives/fmriprep/sub-HA01/func/sub-HA01_task-painexpectancy_run-03_from-boldref_to-T1w_mode-image_desc-coreg_xfm.txt')
t1_to_mni = nitransforms.manip.load('derivatives/fmriprep/sub-HA01/anat/sub-HA01_from-T1w_to-MNI152NLin2009cAsym_mode-image_xfm.h5', fmt='h5')

boldref_to_mni = boldref_to_t1 + t1_to_mni

img_mni = nitransforms.resampling.apply(boldref_to_mni, img, reference=mask)
```

but it crashed with 

```
File .../.env/lib/python3.10/site-packages/nitransforms/resampling.py:122, in _apply_serial(data, spatialimage, targets, transform, ref_ndim
, ref_ndcoords, n_resamplings, output, input_dtype, order, mode, cval, prefilter, max_concurrent)
    112 for t in range(n_resamplings):
    113     xfm_t = (
    114         transform if (n_resamplings == 1 or transform.ndim < 4) else transform[t]
    115     )
    117     targets_t = (
    118         ImageGrid(spatialimage).index(
    119             _as_homogeneous(xfm_t.map(ref_ndcoords), dim=ref_ndim)
    120         )
    121         if targets is None
--> 122         else targets[t, ...]
    123     )
    125     data_t = (
    126         data
    127         if data is not None
    128         else spatialimage.dataobj[..., t].astype(input_dtype, copy=False)
    129     )
    131     tasks.append(
    132         asyncio.create_task(
    133             worker(
   (...)
    146         )
    147     )

IndexError: index 1 is out of bounds for axis 0 with size 1
```

Reverted to `24.1.2` before #247 from @oesteban , it runs for quite a long time, but completes. 

Am I doing it wrong for 3d resampling of 4d data?

mgxd:
Fixes #297 

The interpolation coordinates in `apply()` differed depending on whether the transform was 3D or 4D, and each resampling path handled their shape differently. This unifies the `targets` structure to be `(number of transforms, number of dimensions, number of voxels)`. This ends up fixing a few issues:

- Single 3D transform w/ 4D data crashed.
- 4D transform w/ single 3D volume: returned incorrect data in the non-serialized path.


nitransforms/tests/test_resampling.py
```diff
@@ -376,26 +376,85 @@ def test_LinearTransformsMapping_apply(
...
+
+    mat = np.eye(4)
+    mat[0, 3] = 1  # single shift broadcast over all volumes
+    expected = [((8, 2, 2), vol + 1) for vol in range(nvols)]
+    _assert_apply(img, nitl.Affine(mat), expected, **kwargs)
+
+
+@pytest.mark.parametrize("serialize_4d", [True, False])
+def test_apply_4d_transform_3d_data(serialize_4d):
+    """Regression test for a 4D transform applied to a single 3D volume."""
```
effigies: I personally would be hesitant to apply a 4D transform to a 3D image. I think it's rarely a thing someone wants to do, so doing it silently is more likely to be a bug than a feature. If that is being introduced here, I would say we shouldn't do it, and that the caller should explicitly broadcast the image. If we already support it, then there's nothing to object to.

effigies (approved):
Tests LGTM, and it fixes the issue. One question on API.

nitransforms/tests/test_resampling.py
mgxd: there was never a guard against it, so it was possible, but was being done incorrectly anyways. Feels reasonable to me to just error.

mgxd:
nvm, we actively do use it:

```
___________________ test_LinearTransformsMapping_apply[True] ___________________
[gw0] linux -- Python 3.13.15 /opt/conda/envs/nitransforms/bin/python

tmp_path = PosixPath('/tmp/pytest-of-unknown/pytest-0/popen-gw0/test_LinearTransformsMapping_a0')
data_path = PosixPath('/src/nitransforms/nitransforms/tests/data')
testdata_path = PosixPath('/data'), serialize_4d = True

    @pytest.mark.parametrize("serialize_4d", [True, False])
    def test_LinearTransformsMapping_apply(
        tmp_path, data_path, testdata_path, serialize_4d
    ):
        """Apply transform mappings."""
        hmc = nitl.load(
            data_path / "hmc-itk.tfm", fmt="itk", reference=testdata_path / "sbref.nii.gz"
        )
        assert isinstance(hmc, nitl.LinearTransformsMapping)
    
        # Test-case: realign functional data on to sbref
        nii = apply(
            hmc,
            testdata_path / "func.nii.gz",
            order=1,
            reference=testdata_path / "sbref.nii.gz",
            serialize_nvols=2 if serialize_4d else np.inf,
        )
        assert nii.dataobj.shape[-1] == len(hmc)
    
        # Test-case: write out a fieldmap moved with head
        hmcinv = nitl.LinearTransformsMapping(
            np.linalg.inv(hmc.matrix), reference=testdata_path / "func.nii.gz"
        )
    
>       nii = apply(
            hmcinv,
            testdata_path / "fmap.nii.gz",
            order=1,
            serialize_nvols=2 if serialize_4d else np.inf,
        )

/src/nitransforms/nitransforms/tests/test_resampling.py:358: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

transform = array([[[ 9.99998116e-01, -1.76968320e-03, -7.74861518e-04,
         -8.68600034e-03],
        [ 1.76819976e-03,  9.99...01,
          2.83884096e-01],
        [ 0.00000000e+00,  0.00000000e+00,  0.00000000e+00,
          1.00000000e+00]]])
spatialimage = <nibabel.nifti1.Nifti1Image object at 0x7fa5c64b9310>
reference = None, order = 1, mode = 'constant', cval = 0.0, prefilter = True
output_dtype = None, dtype_width = 8, serialize_nvols = 2, max_concurrent = 2

    def apply(
        transform: TransformBase,
        spatialimage: Union[str, Path, SpatialImage],
        reference: Union[str, Path, SpatialImage] = None,
        order: int = 3,
        mode: str = "constant",
        cval: float = 0.0,
        prefilter: bool = True,
        output_dtype: np.dtype = None,
        dtype_width: int = 8,
        serialize_nvols: int = SERIALIZE_VOLUME_WINDOW_WIDTH,
        max_concurrent: int = min(cpu_count(), 12),
    ) -> Union[SpatialImage, np.ndarray]:
        """
        Apply a transformation to an image, resampling on the reference spatial object.
    
        Parameters
        ----------
        transform: :obj:`~nitransforms.base.TransformBase`
            The 3D, 3D+t, or 4D transform through which data will be resampled.
        spatialimage : :obj:`~nibabel.spatialimages.SpatialImage` or :obj:`os.PathLike`
            The image object containing the data to be resampled in reference
            space
        reference : :obj:`~nibabel.spatialimages.SpatialImage` or :obj:`os.PathLike`
            The image, surface, or combination thereof containing the coordinates
            of samples that will be sampled.
        order : :obj:`int`, optional
            The order of the spline interpolation, default is 3.
            The order has to be in the range 0-5.
        mode : :obj:`str`, optional
            Determines how the input image is extended when the resamplings overflows
            a border. One of ``'constant'``, ``'reflect'``, ``'nearest'``, ``'mirror'``,
            or ``'wrap'``. Default is ``'constant'``.
        cval : :obj:`float`, optional
            Constant value for ``mode='constant'``. Default is 0.0.
        prefilter : :obj:`bool`, optional
            Determines if the image's data array is prefiltered with
            a spline filter before interpolation. The default is ``True``,
            which will create a temporary *float64* array of filtered values
            if *order > 1*. If setting this to ``False``, the output will be
            slightly blurred if *order > 1*, unless the input is prefiltered,
            i.e. it is the result of calling the spline filter on the original
            input.
        output_dtype : :obj:`~numpy.dtype`, optional
            The dtype of the returned array or image, if specified.
            If ``None``, the default behavior is to use the effective dtype of
            the input image. If slope and/or intercept are defined, the effective
            dtype is float64, otherwise it is equivalent to the input image's
            ``get_data_dtype()`` (on-disk type).
            If ``reference`` is defined, then the return value is an image, with
            a data array of the effective dtype but with the on-disk dtype set to
            the input image's on-disk dtype.
        dtype_width : :obj:`int`
            Cap the width of the input data type to the given number of bytes.
            This argument is intended to work as a way to implement lower memory
            requirements in resampling.
        serialize_nvols : :obj:`int`
            Minimum number of volumes in a 3D+t (that is, a series of 3D transformations
            independent in time) to resample on a one-by-one basis.
            Serialized resampling can be executed concurrently (parallelized) with
            the argument ``max_concurrent``.
        max_concurrent : :obj:`int`
            Maximum number of 3D resamplings to be executed concurrently.
    
        Returns
        -------
        resampled : :obj:`~nibabel.spatialimages.SpatialImage` or :obj:`~numpy.ndarray`
            The data imaged after resampling to reference space.
    
        """
        if reference is not None and isinstance(reference, (str, Path)):
            reference = _nbload(str(reference))
    
        _ref = (
            transform.reference
            if reference is None
            else SpatialReference.factory(reference)
        )
    
        if _ref is None:
            raise TransformError("Cannot apply transform without reference")
    
        singleton_4d = spatialimage.ndim == 4 and spatialimage.shape[-1] == 1
        spatialimage = squeeze_image(spatialimage)
    
        # Avoid opening the data array just yet
        input_dtype = cap_dtype(get_obj_dtype(spatialimage.dataobj), dtype_width)
    
        # Number of data volumes
        data_nvols = 1 if spatialimage.ndim < 4 else spatialimage.shape[-1]
        # Number of transforms: transforms chains (e.g., affine + field, are a single transform)
        xfm_nvols = 1 if transform.ndim < 4 else len(transform)
    
        if xfm_nvols > 1 and data_nvols != xfm_nvols:
>           raise ValueError(
                f"A 4D transform ({xfm_nvols} volumes) can only be applied to 4D "
                f"data with a matching number of volumes (got {data_nvols})."
            )
E           ValueError: A 4D transform (8 volumes) can only be applied to 4D data with a matching number of volumes (got 1).

/src/nitransforms/nitransforms/resampling.py:232: ValueError
```

mgxd:
@bpinsard do you mind testing this patch on your example?

bpinsard:
Just tested it, it works. 
Thanks for the super-fast fix!

